PR #14, opened 2023-01-26
Fix start -> star typo in README

XenHat:
I believe this is the original intent


johnfanv2:
Great, you are correct! Thanks a lot for the contribution.

PR #24, opened 2023-03-09
Create PKGBUILD

MrDuartePT:
I create PKGBUILD for arch i will put in the AUR.
I dont have problems in maintain myself.

### DIFF
--- a/deploy/PKGBUILD/LenovoLegionLinux.install
+++ b/deploy/PKGBUILD/LenovoLegionLinux.install
@@ -1,0 +1,24 @@
+post_install() {
+	depmod -a $(uname -r)
+	rmmod legion-laptop.ko #unload old module
+	insmod /usr/lib/modules/$(uname -r)/kernel/drivers/platform/x86/legion-laptop.ko force=1 #force load module
+	echo Installation Finished
+	#Verify if module loaded
+	if [[ $(dmesg | grep "legion_laptop loaded for this device") ]]; then 
+		echo "legion_laptop module loaded"
+	else
+		echo "Module dosent load reboot the machine or verify if your legion is compatible"
+		echo "More info: https://github.com/johnfanv2/LenovoLegionLinux"
+	fi
+}
+
+post_upgrade() {
+ post_install $1
+}
+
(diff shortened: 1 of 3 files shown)


johnfanv2:
Thanks a lot @MrDuartePT. This looks great. I am not sure how to best install the Python stuff in Arch. I already tried a few things on Ubuntu.

- I already created a Python package. There is a install script that uses the python package. https://github.com/johnfanv2/LenovoLegionLinux/blob/main/deploy/build_python.sh. But I guess it might be better to manage all these files in the Arch way.
- There is also a .desktop file. The GUI has to run as root, so maybe start it with pkexec, so I used pkexec. It also needs a policy file to run a GUI app. Currently, both files assume the Python package, so it starts without ".py" extension. I guess that has to change if you do not want to use the python package. But I would be happy if you change it and confirm that it works.

MrDuartePT:
The arch way is particually the same way that use pip probably I just to do a few changes to your script or nothing at all

johnfanv2:
Good idea.

MrDuartePT:
Sorry for the mess now its working you can merge

MrDuartePT:
I detect that the Lenovo_gui module is not installing correctly I will see tomorrow I don't have head for that today.

MrDuartePT:
Tomorrow I will implement only the reload module after isntall and after I will try to solve the Python problem.

MrDuartePT:
Now is full working including the python gui and cli app

MrDuartePT:
Now is full working i have to use pip when installing to force to install the dependencies now i just have to see if will be accept on the AUR.

image

image

Note: the .dektop file work on X11 on Wayland is problem with my xhost configuration

johnfanv2:
Thanks for all the packaging work. I added you as a collaborator to the repository. I assume that you have to accept it. I hope that you can then merge it yourself. I added you because I guess you would have to do small fixes until it works in the AUR. And I will not have much time the next weeks to always merge small changes. Please let me know if something does not work.

I am not sure if you have to create a master branch for AUR or if it works with main.

PR #49, opened 2023-04-28
Added dkms install/remove;fixes

msmafra:
Added dkms install/remove;
Privileged elevation logic fixed;
Updated usage;

### DIFF
--- a/kernel_module/patch_kernel.sh
+++ b/kernel_module/patch_kernel.sh
@@ -1,29 +1,34 @@
 #!/usr/bin/env bash
-# From the discussion https://github.com/johnfanv2/LenovoLegionLinux/pull/29
 ##########################################
-# Checks for doas, sudo, bat and cat first
-# Runs make, make reloadmodule and make install.
+# Based on the script made by @ARYN-27
+# From the discussion https://github.com/johnfanv2/LenovoLegionLinux/pull/29
+# Checks for doas, sudo, bat, dkms and cat first
+# Runs make, make reloadmodule, make install, make uninstall and dkms.
 # Also reloads/restarts power-profiles-daemon when done.
 # Test mode and Install mode
 ##########################################
-lll_kernel_patching()
-                      {
+lll_kernel_patching() {
     LC_ALL=C
     LANG=C
 
     local script_folder
+    local package_name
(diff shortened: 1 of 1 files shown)


johnfanv2:
The DKMS install and all other commands are already inside the make file in the same folder. Can you move your definitions inside the makefile (without power-profile-deamon) to make it consistent?

msmafra:
> The DKMS install and all other commands are already inside the make file in the same folder. Can you move your definitions inside the makefile (without power-profile-deamon) to make it consistent?

I actually based on the documentation and from this one LenovoLegionLinux/deploy/build-dkms.sh with the intention to add it to the README.md.
I can replicate it on the MakeFile, but I don't know how it works. Will it accept any bash built-in commands besides system ones?

johnfanv2:
The file build-dkms.sh will be removed but has not not deleted yet. The installation of dkms is already in the makefile. You might want to call it from your script. You can run commands in the the makefile like in bash. However, note that each line is usually run in a new bash shell. If you do not want that you can "continue" the line with `\` (see Makefile).

msmafra:
Are there more files that will be removed?
I have started some changes on many of the bash scripts inside the **deploy** and **test** folders. Also, I'm trying to figure out a way to make the _setmyfancurve.sh_ easier (in lack of a better word) to use for testing fan curves.

johnfanv2:
The scripts in deploy are currently only used in the CI pipeline (=github actions). They are scripts and not just commands in the github actions so that they can also be run locally for testing when developing. Note that e.g. the script `load.sh` loads it 20 times. So these are really for testing .

What are your ideas?

msmafra:
Hello
So I'll just add 'sudo/doas make DKMS to the script' and keep the remove DKMS (should I?).
On the case of the other scripts, I'll roll them back.
What I had in mind is making the usage of the setmyfancurve.sh easier, a CLI of sorts, and also try to make SecureBoot easy to enable via script.
BUT, on the case of SecureBoot at least, I can't test because my laptop is not enabling it (trying to make Lenovo solve it, they replace my motherboard and now SecureBoot is not working). The setmyfancurve.sh I'll take a while to get my head around it to make it simple to use.

PR #57, opened 2023-05-09
Using make dkms instead; changes to make file and README.md

msmafra:
Replace the dkms block with sudo/doas make dkms;
Kept remove dkms;
Made changes to kernel_module/Makefile to be similar to what was on patch_kernel.sh dkms section;
Correct README.md where Ctrl+Q as being referred as Ctr+Q and added Fn+Q where needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ echo 1800 > /sys/module/legion_laptop/drivers/platform:legion/PNP0C09:00/hwmon/h
 cat /sys/kernel/debug/legion/fancurve
 ```
 Expected: 
-- the controller might have loaded default values if you pressed Ctr+Q(or FN+Q on certain devices) to change the power mode or waited too long; then try again
+- the controller might have loaded default values if you pressed Ctrl+Q(or FN+Q on certain devices) to change the power mode or waited too long; then try again
 - The entries in the fan curve are set to their values. The other values are not relevant (marked with XXXX)
 ```
 rpm1|rpm2|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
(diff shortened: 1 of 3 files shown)


johnfanv2:
Looks good. Thanks a lot for the contribution and putting it into the Makefile.

Regarding your comments on the other pull request:

> So I'll just add 'sudo/doas make DKMS to the script' and keep the remove DKMS (should I?).
> On the case of the other scripts, I'll roll them back.
Are you talking about the patch_kernel.sh and Makefile. This looks good. I am not exactly sure what you mean? Could you give some details.
> What I had in mind is making the usage of the setmyfancurve.sh easier, a CLI of sorts, and also try to make SecureBoot easy to enable via script.
> BUT, on the case of SecureBoot at least, I can't test because my laptop is not enabling it (trying to make Lenovo solve it, they replace my motherboard and now SecureBoot is not working). The setmyfancurve.sh I'll take a while to get my head around it to make it simple to use.
Currenlty, `setmyfancurve.sh` is expected to be edited by the user to set the fan curve and works kind of a example. What do you plan here? Maybe I can give some feedback before you try to modify it.
Currently, there is no script to make it work with secure SecureBoot. It would be great if there is one, but if you can't test it, then I guess it is not a good idea to make one.

msmafra:
> Are you talking about the patch_kernel.sh and Makefile. This looks good. I am not exactly sure what you mean? Could you give some details.

Talking about the ones in the deploy folder. 

> Currenlty, `setmyfancurve.sh` is expected to be edited by the user to set the fan curve and works kind of a example. What do you plan here? Maybe I can give some feedback before you try to modify it. 

The Ideas would be making a config file with the curves and/or pass them via command line, something like:

```
setmyfancurve.sh fan 1500 1700 2100 2500 2700 3000 3500, 1900 2100 2400 2700 3000 3300 3500
setmyfancurve.sh temp 0 46 52 56 60 67, 50 59 65 73 80 85
setmyfancurve.sh -f fancurve.conf
```
> Currently, there is no script to make it work with secure SecureBoot. It would be great if there is one, but if you can't test it, then I guess it is not a good idea to make one.
SecureBoot I have to solve it with Lenovo support, because I also want to keep it enabled from now on. Just when this is solved I try to test it/make a script.

PR #64, opened 2023-06-06
Fix DKMS build fail when updating kernel

Kiblyn11:
There is a variable name mismatch between dkms.conf and Makefile.
When updating kernel, DKMS try to build on old kernel which does not exist.


johnfanv2:
What command did you use to get this errror? Was it `make dkms`? Then I can add a test such that it will no break in the future.

johnfanv2:
Great catch! Thanks for debugging it and fixing it.

PR #89, opened 2023-07-28
kernel_module: wwarning: reset the text color

hraj9258:


MrDuartePT:
I know it a small change but let just the workflow run.

PR #98, opened 2023-08-23
Add dark and light monochromatic logos

rozwell:
Slightly modified and optimized to be used as taskbar icons


rozwell:
@MrDuartePT  check them out if these are ok or should I adjust something?

rozwell:
There should be a setting to choose one or another for the taskbar icon.
Maybe with automatic system theme detection which would initially choose a matching one?

Furthermore, system menu icon should be theme specific.

MrDuartePT:
I just need to see a way to make the python app detect if using dark or light mode before accepting

MrDuartePT:
I found a way to detect the theme and change the I just need to update everthing including the python package dependency and tomorrow update the code and accept you PR

MrDuartePT:
Done everthing can be merge now @johnfanv2 just need you to give a look to my latest commit just to solve a problem 
I tried to allow to set the old tay icon and failed to do it must most of the code is written (comment out) i thing is something with LegionController Class

PR #104, opened 2023-09-01
Swap the dark and light logos

brainos233:
Swap `legion_logo_dark.png` and `legion_logo_light.png`, because the white icon is more suitable for the dark theme, while the black icon is more suitable for the light theme.

<details><summary>Before change</summary>
<p>

image

</p>
</details> 

<details><summary>After change</summary>
<p>

image

</p>
</details>


rozwell:
IMHO it's a code typo, because I named these icons based on what they are, not in which theme they will be used ;)

If theme is supposed to be the main determinant, then we shall rename these logos to something like `dark_theme_logo.png` & `light_theme_logo.png` to make more sense.

brainos233:
> IMHO it's a code typo, because I named these icons based on what they are, not in which theme they will be used ;)
> 
> If theme is supposed to be the main determinant, then we shall rename these logos to something like `dark_theme_logo.png` & `light_theme_logo.png` to make more sense.

Well, I think using `dark`, `light` as a suffix is better than a prefix.

rozwell:
> Well, I think using `dark`, `light` as a suffix is better than a prefix.
Minor semantics ;)

johnfanv2:
I guess these are good points. This definitely needs a change. The only problem I see that in standard Ubuntu/Gnome the system tray is black in both dark mode and light mode.

What is the color of the the system tray on your distros in both modes?

rozwell:
> I guess these are good points. This definitely needs a change. The only problem I see is that in standard Ubuntu/Gnome the system tray is black in both dark mode and light mode.
> 
> What is the color of the the system tray on your distros in both modes?

On Kubuntu it's all normal here:
- light theme uses dark icons
- dark theme uses light icons

MrDuartePT:
ok i will do the change and all create a new release

MrDuartePT:
This create problem on dark background I will revert this

rozwell:
@MrDuartePT is there an option to manually switch between these icons?

MrDuartePT:
I think not @johnfanv2 how much is difficult to add that?

MrDuartePT:
I will revert this for now just to not break in other system

rozwell:
@johnfanv2 let us know if & when you'll start working on this, cause I might have a little time to do that myself ;)

MrDuartePT:
@rozwell you can make and submit a PR

MrDuartePT:
I gonna verify one thing to see if a really need to revert this.

rozwell:
> The only problem I see is that in standard Ubuntu/Gnome the system tray is black in both dark mode and light mode.

I initially made white icon with a little shadow so it would work well on any background, but these icons are to small for that and it was not looking good.


> @rozwell you can make and submit a PR

I can once I'll have some time to spare, but that could take a few days 😐

MrDuartePT:
No problem😀

MrDuartePT:
I found out the problem in Kubuntu and Ubuntu the tray are dark but the system GTK theme is set to light @brainos233 make a good observation our system the way there are setup is not right

johnfanv2:
You can now set the color of the logo in the settings to a fixed scheme or change automatically. Please check if it works on your distro.

MrDuartePT:
I can confirm it works

PR #105, opened 2023-09-02
Revert "Swap the dark and light logos"

MrDuartePT:
This make logos not visible in dark modes that use dark tray (most common type of dark mode)


MrDuartePT:
@brainos233 i think you kde is set to light mode but your dock is grey that why is not visible.

PR #120, opened 2023-09-10
Add missing second slashes in the package_repo

Abrifq:


Rittmeister39:
thanks @Abrifq ! I suppose I need to wait now until maintainer allows this fix into the repo. then i can install it only, right?

Abrifq:
> I suppose I need to wait now until maintainer allows this fix into the repo. then i can install it only, right?

You can install it directly via the deb files until maintainer accepts this fix.

Get both lenovolegionlinux and darkdetect packages files and install them locallu.

Rittmeister39:
thanks. I try that later.

MrDuartePT:
Well i just need to verify if this file is change when the rpm and deb build if that is the case I need to fix that but thanks for finding this typos

MrDuartePT:
I will accept this but i will also fix the workflow file since is where the problem is.

PR #129, opened 2023-10-16
fix-directory-in-fancurve-set

BoskyWSMFN:
* fixed annoying "...//...: No such file or directory".


MrDuartePT:
this is not a error you didnt copy /usr/share/legion-linux to /etc/legion-linux that why you are getting no such file or directory.
image

PR #130, opened 2023-10-17
Improve the README, add power profile support for 15ARH05H

MasterKia:
Power profile does work on 15ARH05H.


README.md
```diff
@@ -809,175 +809,180 @@ If otherwise the fans never stop, you might have set a very low upper temperatur
...
+- ~~As far as I know, power saving or performance of CPU does not change, as this is controlled by kernel or tools like cpupower~~
+- An analysis of the ACPI firmware, which is "stored in hardware", shows that some power options, e.g. the STAPM limit, of the CPU and GPU are changed via ACPI calls. This can be observed by changing power mode with Fn+Q and checking CPU power limits, e.g. with `ryzenadj` for AMD CPUs. ~~However, no changes to the GPU power limits are observed on NVIDIA GPUs, which might indicate that the integration of ACPI and GPU driver is faulty on Linux~~ On driver 535+, GPU power limits for nvidia changes on battery power (40w on quiet and 50w on balanced for RTX 3070), on power adapter, the power limit are allways set to max on peformance (reported by `nvidia-smi`), but more power is allowed to the GPU on performance or custom mode in demanding GPU tasks. Currently the power limit on performance mode is stuck on 80W compared to 115W/130W on Windows (see: https://forums.developer.nvidia.com/t/lenovo-legion-5-rtx-3060-460-x-470-x-495-x-power-draw-stuck-at-80w-windows-does-130w/192174).
```
MrDuartePT, line 856: The last part you add about the power stuck to 80w I get 120w on performance and 140w in custom mode when change the gpu power values in the gui (need to update the readme). To solve the problem on getting stuck at 80w you need to enable nvidia-powerd.service

MrDuartePT (changes requested):
See the comment I put in the code and please changing. Besides that nice work.

README.md
MrDuartePT: Well depend of the drive on 535 work fine for me and other with amd system.

MrDuartePT (approved):
Look good. Thanks

README.md
MrDuartePT: Strange for me work but I have a gen 6 this have ryzen 5000 instead of 4000 can be that

MrDuartePT: > > I would recommed your to try the lastest driver
> 
> I'm already on `535.113.01`.
> 
> > if dosent work rewrite like that.
> 
> Done.
> 
> > send the bug here to here I gonna have a look
> 
> Please see: [NVIDIA/open-gpu-kernel-modules#483](https://github.com/NVIDIA/open-gpu-kernel-modules/issues/483)

This make impossible change the power limit manually but the gpu will boost to the highest tdp on performance mode. And in custom you can change the boost values in the gui.

MrDuartePT: > `nvidia-smi --query-gpu=power.max_limit --format=csv` returns:
> 
> ```
> power.max_limit [W]
> 115.00 W
> ```
> 
> But if I try to change the power limit it doesn't work, `sudo nvidia-smi -i 0 -pl 115`:
> 
> ```
> Changing power management limit is not supported for GPU: 00000000:01:00.0.
> ```

New driver make impossible to change the value for one set by the user on laptop but the card will still boost in performance and in custom mode you can change the boost values.

MrDuartePT: Probably you having problem with nvidia powerd because I think you are using Optimus manager.

MrDuartePT: If you go to the discord server we can try help you with that. 
https://discord.gg/legionseries

MrDuartePT: Ok jus unistall Optimus manager (set on hybrid) and use superglfxctl

PR #134, opened 2023-10-25
use `size_t` instead of `int` when iterating through `fancurve->points`

neeythann:
While `MAXFANCURVESIZE` wont be greater than `MAX_INT`, it's much better to use the same type of `fancurve->size` when iterating.


neeythann:
My bad, I've re-read the code base again and it seems that this is not needed.

PR #155, opened 2024-01-15
fix PyQt6 hide to tray

st0nie:


st0nie:
Otherwise, if you enable minimize when opening the software, the software will immediately flash back and won't work!

PR #168, opened 2024-02-19
Wayland CI fix v2

MrDuartePT:
This is for CI fixing will not change nothing for the user.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,9 @@ jobs:
       - name: Install Development Dependencies
         run: ./deploy/dependencies/install_development_dependencies_ubuntu.sh
         shell: bash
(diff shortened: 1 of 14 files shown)


MrDuartePT:
This is a simplified version using xwayland-run project. Is failing problaby because is still in develoment

st0nie:
Thanks for trying to improve the GUI CI. The workflow has changed quite a bit since this draft was opened, and this branch no longer matches current `main`.

It also has a `menos`/`meson` typo and depends on cloning an unpinned external project during CI. I think it would be better to close this draft and, if Wayland GUI CI is still needed, open a fresh minimal PR against the current Ubuntu 24.04 workflow using packaged or pinned dependencies.

PR #176, opened 2024-03-06
Create legion-linux-onresume.service

st0nie:
Some modifications, such as the changes made by ryzenadj, cannot be preserved after the machine goes through the suspend-resume cycle. Therefore, we can create a file to allow the Legion service to automatically start once the machine resumes.

### DIFF
--- a/extra/service/legion-linux-onresume.service
+++ b/extra/service/legion-linux-onresume.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Restart legion-linux on resume
+Before=sleep.target
+StopWhenUnneeded=yes
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/bin/systemctl restart legion-linux.service
+TimeoutStopSec=infinity
+
+[Install]
+WantedBy=sleep.target


st0nie:
`sudo mv this_file_path /usr/lib/systemd/system-sleep/legion` to make it works
require systemd, of course

st0nie:
Now I create `legion-linux-onresume.service` and it seems better
Just enable the service,legion-linux will auto restart on system resume

st0nie:
The reason why I use `restart` instead of `try-restart` is that `legion-linux.service` will be inactive after a while, and `try-restart` only restart service which is `active`

st0nie:
cc
@MrDuartePT

MrDuartePT:
That why I noticed some times systemd is was degraded and the service failed to start.
Thanks to find a solution I will just need to do some changes to the packaging.

st0nie:
> That why I noticed some times systemd was degraded and the service failed to start. Thanks to find a solution I will just need to do some changes to the packaging.

IDK whether the fancurve-set will preserve its modification
But  changes made by ryzenadj will 100% lost after resume

MrDuartePT:
I giving a look now and my service was failing because I change to a kernel with the kernel module install.
This seems to be a bug only with ryzenadj.

Well I also thinking of use the legion_cli.py as deamon instead of a seperate bash script, but we can add this systemd service in the mid time.
The ryzenadj support just exist because some of cpu limit stuff in the gui dosen't work in some older model, so user can use ryzenadj to also set cpu option (and ryzenadj have more option), but it can be edit to other tool if you are using intel.

st0nie:
> Well I also thinking of use the legion_cli.py as deamon instead of a seperate bash script, but we can add this systemd 

`legion-linux.service` as a daemon is probably better than now
Now this is a script and we cannot make sure it is really working after we exec it.
If it is a daemon, we can do some check after `/sys/firmware/acpi/platform_profile` changed and or  pluged off the AC
After all, daemon is simpler for user, they just need to enable the service, and `legion-linux daemon` do all the things

MrDuartePT:
just a quick note also work in `/usr/lib/systemd/system/legion-linux-onresume.service`

PR #180, opened 2024-03-11
Migrate to legiond.service

### LINKED ISSUE #179: Use timer and unix_socket to run legion-linux.service
[legion.pdf](https://github.com/johnfanv2/LenovoLegionLinux/files/14553464/legion.pdf)

CC @MrDuartePT 

It will solve some problems such as run legion-linux.service freqently and hit the limit set by systemd
And resume from suspend and AC adapter will be easier to run fancurve-set (racing free)

For example, a plasma user config plasma  -- setting `ppd` to powersaver when AC plug off, and his previous profile is powersaver

## With the original model
1. legion.path not run fancurve-set
2. acpid want to run fancurve-set

### When he plug in the AC
1.legion.path want to run fancurve-set
2.acpid want to run fancurve-set

Obviously here is a race running
### plug off:
1.legion.path not send request
2.acpid send request to legiond

after 3s , fancurve-set is done

### plug in:
1.legion.path send request, timer create
2.acpid send request to legiond , timer reset

after 3s , fancurve-set is done

## I found this race condition Cuz I noticed thats after I plug in my AC adapter, my `ryzenadj` modifycation lost

st0nie:
close #179

Replace legion.path with `legion_ppd` acpid event to prevent systemd's limit
Replace legion-linux.service with legiond

## usage
compile legiond and legiond-cli via `make`
Make sure `legiond` and `legiond-cli` are in `/usr/bin/`
Then
``` sh
systemctl disable legion-linux*
systemctl enable --now legiond.service legiond-onresume.service
cp legion_ac /etc/acpi/events/legion_ac
cp legion_ppd /etc/acpi/events/legion_ppd
```

Enjoy it!

Tested on my machine and works nearly perfect


st0nie:
cc @MrDuartePT

st0nie:
seems good enough @MrDuartePT

MrDuartePT:
I just don‘t know if is possible to package python and c compile code, need to check that.

Also don't know if @johnfanv2 would prefer that to be implemented in the python program

st0nie:
> Also don't know if @johnfanv2 would prefer that to be implemented in the python program

I feel that this program is already sufficiently concise, so there may not be a need to reimplement it in Python. C language offers superior performance, which is highly important for a daemon.

Compiling and installing them is also straightforward - you just need to run make.

st0nie:
Anyway，Plz try it first
I feel it comfortable to use.

MrDuartePT:
> Anyway，Plz try it first I feel it comfortable to use.

I will try it not worry

MrDuartePT:
@st0nie since nvdia driver not allow tdp control and since it exist know way to edit or create systemd service for ryzenadj or other tools. 

I think it better to only give this deamon the fan control profile loading, since I pdon't know if people use it for apply ryzenadj setting

st0nie:
> I think it better to only give this deamon the fan control profile loading, since I don't know if people use it for apply ryzenadj setting

I don't think this is important because these are indeed changes that need to be made after the platform_profile is altered. Once the platform_profile is changed, the fan curve will be restored to default, and the same applies to ryzenadj. Therefore, it makes sense to apply these changes simultaneously. There is no issue with the fancurve-set script; the problem lies in the way and frequency of launching fancurve-set.

While we can add integration for tools like ryzenadj to legiond, it may not be necessary to do so, and it might be better to keep the original fancurve-set functionality as it is.

MrDuartePT:
> > I think it better to only give this deamon the fan control profile loading, since I don't know if people use it for apply ryzenadj setting
> 
> 
> 
> I don't think this is important because these are indeed changes that need to be made after the platform_profile is altered. Once the platform_profile is changed, the fan curve will be restored to default, and the same applies to ryzenadj. Therefore, it makes sense to apply these changes simultaneously. There is no issue with the fancurve-set script; the problem lies in the way and frequency of launching fancurve-set.
> 
> 
> 
> While we can add integration for tools like ryzenadj to legiond, it may not be necessary to do so, and it might be better to keep the original fancurve-set functionality as it is.

Ok it makes sense this way if someone decides can also alter the script in a way that have more functionality

MrDuartePT:
Let me test it, warn me if you going to do any change

MrDuartePT:
works really good, just require acpid to be install. I gonna accept this and update the packages.

PR #187, opened 2024-03-22
Add legiond-cpuset.{service,timer}

### LINKED ISSUE #186: introduce cpu reset timer support
After testing, even if the configuration of ryzenadj was applied correctly at the beginning, after keeping the computer turned on and idle for a while, it will still reset, ryzen-ppd will reset the configuration every 2s to make sure that the configuration is valid, while undervolt's timer is every 30s

st0nie:
close #186 
https://github.com/georgewhewell/undervolt
https://github.com/xsmile/ryzen-ppd/

Referenced the link above

Introduce A timer so user can choose to enable it to make cpu tweaks always work
Otherwise CPU changes may revert after a period of time

CC @MrDuartePT

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,7 +708,9 @@ This systemd service also have extras features that can be activated by editing 
     - GPU_TDP_AC_P - Custom GPU TDP for performance mode on charger
   - Note: The default values in the .env file are from RTX 3070
 
-NOTE: We also have acpi action and event inside the extra folder but are in develoment and testing (use at your own risk).
+NOTE: ~We also have acpi action and event inside the extra folder but are in develoment and testing (use at your own risk).~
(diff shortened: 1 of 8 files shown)


MrDuartePT:
Before accept this I will also do some change like update the Readme. Since this is not documented anywhere how it works😂.

st0nie:
> Before accept this I will also do some change like update the Readme. Since this is not documented anywhere how it works😂.

wait , after tesing ,the fancurve also get reset sometimes……
legiond's modifications are unnecessary.
just add a timer to do the reset per 5 minutes

st0nie:
readme updated

CC @MrDuartePT

MrDuartePT:
I don't use cpu tweak at the moment but I can test it

MrDuartePT:
ok test it and works fine

PR #189, opened 2024-03-30
Optimize gui for high scale

ArcticLampyrid:
Relax max size requirements for main window.
Adjust scroll area, keeping tab headers & bottom buttons fixed.

### DIFF
--- a/python/legion_linux/legion_linux/legion_gui.py
+++ b/python/legion_linux/legion_linux/legion_gui.py
@@ -1063,9 +1063,9 @@ class FanCurveTab(QWidget):
         self.layout.addWidget(self.ic_upper_temp_label, 8, 0)
         self.layout.addWidget(self.accel_label, 9, 0)
         self.layout.addWidget(self.decel_label, 10, 0)
-        self.layout.addWidget(self.minfancurve_check, 11, 0)
-        self.layout.addWidget(self.lockfancontroller_check, 12, 0)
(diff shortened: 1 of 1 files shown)


MrDuartePT:
Well I was doing changes now, but your solution work really well

MrDuartePT:
Just not like to be maximize by default, but the window size is almost perfect, it better to be `725`

python/legion_linux/legion_linux/legion_gui.py
```diff
@@ -1424,9 +1427,8 @@ def __init__(self, controller:LegionController, icon:QtGui.QIcon):
         self.controller = controller
         self.controller.main_window = self
 
-        # Set a Minium Size to the window
-        # Scroll Area make window small square by default
-        self.setMinimumSize(1250, 1000)
```
MrDuartePT, line 1429: Revert this change your make it `self.setMinimumSize(1250, 725)`
image

python/legion_linux/legion_linux/legion_gui.py
```diff
@@ -1717,7 +1712,7 @@ def main():
     # Start Windows
     if controller.model.app_model.automatic_close.get():
         main_window.close_after(3000)
-    main_window.show()
```
MrDuartePT: Not make it Maximized it a bit annoying for people that open the app on login

But for everthing else nice job :)

MrDuartePT (approved):
I push the new changes

PR #190, opened 2024-03-30
Fix typo in comment in legion-laptop.c

makru86:
Correct the spelling of "firmware" in the comment describing the hwmon interface for monitoring fan
speed and temperature and setting the fan curve.


MrDuartePT:
Thanks

PR #200, opened 2024-04-07
Fix kernel module build fail

st0nie:
CC @MrDuartePT @johnfanv2 

I'm not sure if this affects the functionality of the kernel module or not
But it looks like these functions are only called in the same file


MrDuartePT:
@st0nie it shoudn't, test changing powermodes and fancurve in gui (is want this code is responsible to do) if everthing work alright just merge it

PR #215, opened 2024-05-02
Use ubuntu docker container instead of jammy github container

MrDuartePT:


MrDuartePT:
Well no need for this debian have a package: https://tracker.debian.org/pkg/lenovolegionlinux

PR #224, opened 2024-06-03
Docs: Fix Conservation Mode Battery Percentage

Malix-off:


MrDuartePT:
It depend of the model really to be fair, but it dosen't hurt say 60% tho.

Malix-off:
ah yes?
i thought it always was 60%, i just assumed from having tested with 3 lenovo legion, my bad!

MrDuartePT:
> ah yes?
> 
> i thought it always was 60%, i just assumed from having tested with 3 lenovo legion, my bad!

No problem really because the new model they are all hardcoded to 60%.

Issue #250, opened 2024-09-04
[BUG] function signature, failing to compile on kernel-headers-6.11.0-0.rc5.43.fc41.x86_64


imalone:
On Fedora F41 I find this when building:
```
make -C /lib/modules/6.11.0-0.rc5.43.fc41.x86_64/build M=/home/ian/git/LenovoLegionLinux/kernel_module modules
make[1]: Entering directory '/usr/src/kernels/6.11.0-0.rc5.43.fc41.x86_64'
  CC [M]  /home/ian/git/LenovoLegionLinux/kernel_module/legion-laptop.o
/home/ian/git/LenovoLegionLinux/kernel_module/legion-laptop.c:6147:19: error: initialization of ‘void (*)(struct platform_device *)’ from incompatible pointer type ‘int (*)(struct platform_device *)’ [-Wincompatible-pointer-types]
 6147 |         .remove = legion_remove,
      |                   ^~~~~~~~~~~~~
/home/ian/git/LenovoLegionLinux/kernel_module/legion-laptop.c:6147:19: note: (near initialization for ‘legion_driver.<anonymous>.remove’)
/home/ian/git/LenovoLegionLinux/kernel_module/legion-laptop.c:1816:16: warning: ‘ecram_memoryio_write’ defined but not used [-Wunused-function]
 1816 | static ssize_t ecram_memoryio_write(const struct ecram_memoryio *ec_memoryio,
      |                ^~~~~~~~~~~~~~~~~~~~
make[3]: *** [scripts/Makefile.build:244: /home/ian/git/LenovoLegionLinux/kernel_module/legion-laptop.o] Error 1
make[2]: *** [/usr/src/kernels/6.11.0-0.rc5.43.fc41.x86_64/Makefile:1943: /home/ian/git/LenovoLegionLinux/kernel_module] Error 2
make[1]: *** [Makefile:236: __sub-make] Error 2
make[1]: Leaving directory '/usr/src/kernels/6.11.0-0.rc5.43.fc41.x86_64'
make: *** [Makefile:13: all] Error 2
```

Compilation and make reloadmodule tests go okay if I patch the return value of legion_remove:
```
diff --git a/kernel_module/legion-laptop.c b/kernel_module/legion-laptop.c
index 405b198..b510458 100644
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -6084,7 +6084,7 @@ err_model_mismtach:
        return err;
[...]

PR #251, opened 2024-09-07
fix: Make legion_remove return void and set remove_new

johnfanv2:
The type of .remove in platform_driver changed from

int (*)(struct platform_drive*)

to

void (*)(struct platform_drive*)

in kernel 6.11 to force migration. In version before

.remove_new has type

void (*)(struct platform_drive*)

to allows a migration. We use remove_new to make it commpatible with different kernels.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -6084,7 +6084,7 @@ err_model_mismtach:
 	return err;
 }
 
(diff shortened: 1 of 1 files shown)


johnfanv2:
Fixes bug in issue #250.

st0nie (approved):

MrDuartePT (approved):

PR #259, opened 2024-09-16
Add missing missing mkdir command for DKMS installation instructions

### LINKED ISSUE #253: Incorrect cp command for DKMS installation
https://github.com/johnfanv2/LenovoLegionLinux/blob/714cd25a933dcf526149dfa538f3ea6b975df19e/README.md?plain=1#L269

That won't work because target directory doesn't exists:
```
$ sudo cp ./kernel_module/* /usr/src/LenovoLegionLinux-1.0.0 -r 
cp: target '/usr/src/LenovoLegionLinux-1.0.0': No such file or directory
```

We need to either create a new directory first:
```
sudo mkdir /usr/src/LenovoLegionLinux-1.0.0
```
or copy entire directory (without asterisk):
```
sudo cp ./kernel_module/ /usr/src/LenovoLegionLinux-1.0.0 -r 
```

rozwell:
Using `mkdir -p` will not raise an error if directory exists. Fixes #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,7 @@ sudo make uninstall
 You must first install the package to with DKMS. See Requirements section.
 
 ```
+sudo mkdir -p /usr/src/LenovoLegionLinux-1.0.0
 sudo cp ./kernel_module/* /usr/src/LenovoLegionLinux-1.0.0 -r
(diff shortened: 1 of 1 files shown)


johnfanv2:
Good catch!

PR #185, opened 2024-03-19
Add support for the Legion Go

antheas:
This pr adds support for the legion go. Right now, tdp modes and full fan speed work. Fan curves do not, and there is a lot of logging to dmesg.

PR #261, opened 2024-09-22
Convert kernel hwmon interface to use PWM (0-255) intsead of RPM

johnfanv2:
This will make it compliant with Kernel standard for hwmon and compatible with different fan curves (rpm, percentage, ...) of different models.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -1455,7 +1455,7 @@ err:
 //}
 
 static int wmi_exec_ints(const char *guid, u8 instance, u32 method_id,
-			 const struct acpi_buffer *params, u8 *res,
+			 const struct acpi_buffer *params, void *res,
 			 size_t ressize)
 {
 	acpi_status status;
@@ -1518,7 +1518,7 @@ static int wmi_exec_noarg_int(const char *guid, u8 instance, u32 method_id,
 }
 
 static int wmi_exec_noarg_ints(const char *guid, u8 instance, u32 method_id,
-			       u8 *res, size_t ressize)
+			       void *res, size_t ressize)
 {
 	struct acpi_buffer params;
 
@@ -1712,6 +1712,12 @@ enum OtherMethodFeature {
 	OtherMethodFeature_C_U1 = 0x05010000,
 	OtherMethodFeature_TEMP_CPU = 0x05040000,
 	OtherMethodFeature_TEMP_GPU = 0x05050000,
+	OtherMethodFeature_FAN_FULLSPEED = 0x04020000,
+};
+
+struct other_method_val {
+	u32 param;
+	u32 val;
 };
 
 static ssize_t wmi_other_method_get_value(enum OtherMethodFeature feature_id,
@@ -1728,6 +1734,20 @@ static ssize_t wmi_other_method_get_value(enum OtherMethodFeature feature_id,
 			     WMI_METHOD_ID_GET_FEATURE_VALUE, &params, &res);
 	if (!error)
 		*value = res;
+	return error;
+}
+
+static ssize_t wmi_other_method_set_value(enum OtherMethodFeature feature_id,
+					  int value)
+{
+	int error;
+	struct other_method_val param = {
+		.param=feature_id,
+		.val=value
+	};
+
+	error = wmi_exec_arg(LEGION_WMI_LENOVO_OTHER_METHOD_GUID, 0,
+			     WMI_METHOD_ID_SET_FEATURE_VALUE, &param, sizeof(param));
 	return error;
 }
 
@@ -2828,10 +2848,23 @@ struct WMIFanTable {
 	u16 FSS7;
 	u16 FSS8;
 	u16 FSS9;
+	u8 FSSNULL;
+	u32 FTLE;
+	u16 FST0;
+	u16 FST1;
+	u16 FST2;
+	u16 FST3;
+	u16 FST4;
+	u16 FST5;
+	u16 FST6;
+	u16 FST7;
+	u16 FST8;
+	u16 FST9;
+	u8 FSTNULL;
 } __packed;
 
 struct WMIFanTableRead {
-	u32 FSFL;
+	u32 FSTL;
 	u32 FSS0;
 	u32 FSS1;
 	u32 FSS2;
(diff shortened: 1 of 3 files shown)


MrDuartePT:
LGTM, I gonna make the python changes and this can be merge.

#185 need to be update since it have some fixes to new devices

MrDuartePT:
The commit I just push need be tested in a newer device but the code did compile.
The temperature is still hardcoded for the legion go values.

@antheas if you can give a look if I miss something.

antheas:
kernel lgmt 

I would personally keep using rpm for older models and change the driver name though. Perhaps even split the EC fan stuff to a different module that is out-of-tree, and try to merge only the WMI to the kernel.

MrDuartePT:
> I would personally keep using rpm for older models and change the driver name though. Perhaps even split the EC fan stuff to a different module that is out-of-tree, and try to merge only the WMI to the kernel.

Probably you are right and that would need to be done, but since I don't think the drive would be push to mainline anytime soon, let wait.
But I think is one of the things with should ask when making a proposal to kernel mailing list.

kernel_module/legion-laptop.c
```diff
@@ -1712,6 +1712,12 @@ enum OtherMethodFeature {
 	OtherMethodFeature_C_U1 = 0x05010000,
 	OtherMethodFeature_TEMP_CPU = 0x05040000,
 	OtherMethodFeature_TEMP_GPU = 0x05050000,
+	OtherMethodFeature_FAN_FULLSPEED = 0x04020000,
```
johnfanv2, line 1715: What is the source for that? Is there already a function that uses it that is not in this commit? Maybe you already have info how to use it. It is currently not used in this file.

kernel_module/legion-laptop.c
```diff
@@ -2842,84 +2875,93 @@ struct WMIFanTableRead {
...
+	u32 FST0;
+	u32 FST1;
+	u32 FST2;
+	u32 FST3;
+	u32 FST4;
+	u32 FST5;
+	u32 FST6;
+	u32 FST7;
+	u32 FST8;
+	u32 FST9;
```
johnfanv2, line 2892: What is FTLE? Length?
FSTL: is this field renamed in Legion Go?
FST* = temp;
FSS* = speed; is it 0-100 (percent), pwn (0-255), or in rpm?

kernel_module/legion-laptop.c
```diff
@@ -2842,84 +2875,93 @@ struct WMIFanTableRead {
...
 } __packed;
 
 static ssize_t wmi_read_fancurve_custom(const struct model_config *model,
 					struct fancurve *fancurve)
 {
-	u8 buffer[88];
+	struct WMIFanTableRead fantable;
```
johnfanv2, line 2898: Good idea. By adding this new fields we get exactly 88 Bytes. These fields where not defined in the ACPI code of other models.

kernel_module/legion-laptop.c
```diff
@@ -2842,84 +2875,93 @@ struct WMIFanTableRead {
...
-		       DUMP_PREFIX_ADDRESS, 16, 1, buffer, sizeof(buffer),
-		       true);
+		       DUMP_PREFIX_ADDRESS, 16, 1, &fantable,
+		       sizeof(fantable), true);
 	if (!err) {
-		struct WMIFanTableRead *fantable =
-			(struct WMIFanTableRead *)&buffer[0];
 		fancurve->current_point_i = 0;
-		fancurve->size = 10;
+		fancurve->size = fantable.FSTL;
```
johnfanv2, line 2914: Hmm. This does not work for other models because FSTL is not set to 10 but just not set.

kernel_module/legion-laptop.c
```diff
@@ -2842,84 +2875,93 @@ struct WMIFanTableRead {
...
-	buffer[0x16] = fancurve->points[8].speed1;
-	buffer[0x18] = fancurve->points[9].speed1;
-
 	print_hex_dump(KERN_INFO, "legion_laptop fan table wmi write buffer",
-		       DUMP_PREFIX_ADDRESS, 16, 1, buffer, sizeof(buffer),
+		       DUMP_PREFIX_ADDRESS, 16, 1, &fan, sizeof(fan),
 		       true);
 	err = wmi_exec_arg(WMI_GUID_LENOVO_FAN_METHOD, 0,
-			   WMI_METHOD_ID_FAN_SET_TABLE, buffer, sizeof(buffer));
+			   WMI_METHOD_ID_FAN_SET_TABLE, &fan, sizeof(fan));
```
johnfanv2, line 2967: In the old code we write a buffer of size 0x20. Now we write an object of size 0x34. Will this break stuff in old models?

kernel_module/legion-laptop.c
```diff
@@ -5694,6 +5736,34 @@ static struct attribute *fancurve_hwmon_attributes[] = {
 	&sensor_dev_attr_pwm1_mode.dev_attr.attr, NULL
 };
 
+static struct attribute *fancurve_hwmon_attributes_legion_go[] = {
```
johnfanv2, line 5753: Why create new hwmon entries? Do we need new ones with other properties?

kernel_module/legion-laptop.c
```diff
@@ -5694,6 +5736,34 @@ static struct attribute *fancurve_hwmon_attributes[] = {
...
+	&sensor_dev_attr_pwm1_auto_point4_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point5_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point6_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point7_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point8_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point9_pwm.dev_attr.attr,
+	&sensor_dev_attr_pwm1_auto_point10_pwm.dev_attr.attr,
+	// These should be read only if the driver is properly implemented
+	// and either fail to set or noop
+	// &sensor_dev_attr_pwm1_auto_point1_temp.dev_attr.attr,
```
johnfanv2, line 5766: I guess, we can make set read/write properties depending on model or just error out/ignore it.

kernel_module/legion-laptop.c
```diff
@@ -5720,6 +5790,11 @@ static const struct attribute_group legion_hwmon_fancurve_group = {
 	.is_visible = legion_hwmon_is_visible,
 };
 
+static const struct attribute_group legion_go_hwmon_fancurve_group = {
```
johnfanv2, line 5807: See above. I guess, this might be removed.

st0nie:
Thanks for the work that went into this. Since this PR was opened, `main` has gained most of the PWM/RPM and model-aware fan-curve handling, so a large part of this draft is now superseded.

The branch also conflicts with current code and still has unresolved questions around firmware table size, shared WMI payload changes, and Legion Go-specific values in common paths.

I think the cleanest path is to close this old draft and open a small, model-gated PR for any behavior that is still missing on current `main`.

PR #278, opened 2024-11-13
Change beaviour of dracut

bonaccolto90:
It would be better to use 'dracut-rebuild' instead of 'dracut -f --regenerate-all --uefi' because it creates multiple voices for each kernel and, when there is no space left, it will not create the EFI.

Find out I had a couple of EFI voices of old kernel and not the latest.


bonaccolto90:
I realized what that line I removed do... it should be implemented a system that check what version of the kernel is installed and remove the old voices…

PR #294, opened 2025-01-23
fix: fix linux 6.13 build

ToolmanP:
Upstream Commit e70140ba0d2b1a30467d4af6bcfe761327b9ec95 (Get rid of 'remove_new' relic from platform driver struct) changes the struct platform_driver signature, we should apply changes correspondingly but still retain the compatibility for old kernels.

Tested both on 6.12 and 6.13.0 Kernels.


mqasimca:
Hi, @MrDuartePT Can it be merged please

PR #388, opened 2026-01-17
Fix value for speed_pwm2

qquique:
Function fancurve-print-seqfile currently is showing the same values from speed1 in speed2 column. This PR fixes it when calling function fancurve_get_speed_pwm using the corresponding id 1.


MrDuartePT:
Thanks for the PR!

PR #390, opened 2026-01-24
fix CI containers tests

qquique:
- adds archlinux pillow dependency
- updates fedora container version to minimum version supported by package in copr : https://copr.fedorainfracloud.org/coprs/mrduarte/LenovoLegionLinux/
- updates links of package in debian package repository (was moved from contrib to main)
- updates ubuntu container version to 24:04 because packages on updated links have dependency on dkms and glibc versions
- updates ubuntu container pillow dependency


MrDuartePT:
Thanks!

PR #407, opened 2026-03-14
Add support for Legion Pro 7 16IAX10H (Q7CN, Arrow Lake 2025)

nuclearcat:
I recently bought a Legion Pro 7 16IAX10H (83F5, Intel Arrow Lake, BIOS Q7CN) and was impressed to discover this project. After some struggle getting the driver to load on my laptop (the 2025 firmware is quite different from older
  models), I managed to get most core features working.

The main challenge is that Q7CN firmware routes nearly everything through the WMAE (OtherMethod/GetFeatureValue/SetFeatureValue) WMI interface, rather than the dedicated per-feature WMI GUIDs (CPU method, GPU method, fan method) or direct EC register access used by older models. Several generic improvements were needed to support this, which should also benefit other newer models that follow the same pattern.

  What this PR does:
  - Adds WMAE wmi_other_method_set_value() — the write counterpart that was missing
  - Adds ACCESS_METHOD_WMI3 dispatch for fan fullspeed (older models use WMAB, Q7CN uses WMAE 0x04020000)
  - Fixes pwm1_mode hwmon to use the access method dispatch instead of hardcoded EC reads (was broken for any non-EC model)
  - Adds WMAE fallback to CPU/GPU power limit sysfs attributes — tries the dedicated WMI GUID first, falls back to WMAE on failure (backward compatible)
  - Adds no_ylogo_light/no_ioport_light model config flags for models where these LEDs aren't WMI-controllable
  - Adds model_q7cn config and DMI match entry
  - Disables WMI keyboard backlight registration (keyboard is USB HID: ITE 048d:c193 "Lenovo Lighting", i will look into that later)

  What doesn't work / uncertain

  - Fan curves read as all zeros — the WMI3 fan curve read path may need a different approach for this firmware
  - Fan fullspeed — the FNST toggle works (reads back correctly), but I'm not entirely sure it forces unconditional max RPM vs just switching to a more aggressive curve. The fans don't audibly ramp to maximum at idle when enabled, though
  that could be normal EC behavior at low temperatures, i didnt test that well yet.
  - CPU/GPU power limit writes — reads work via WMAE fallback, but writes may need the laptop to be in custom power mode (ODV1=0x03) per the DSDT logic. Needs more testing under load
  - Keyboard backlight — controlled by USB HID (ITE 048d:c193), needs a separate driver?
  - Y-logo / IO-port LEDs — seems not controllable via WMAF on this firmware. The proper control mechanism hasn't been identified yet

Also, I only have this one laptop to test on, so I cannot verify that these changes don't break older Legion models. The generic improvements (WMAE fallback, pwm1_mode fix) are designed to be backward compatible — they try the existing path first and only fall back to WMAE on failure. But please take this with a grain of salt and test on other hardware if possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ It allows you to control features like the fan curve, power mode, power limits, 
 
 # If you have a light in the lid (Y-logo) or at the IO-ports (all Legion 7), please help testing controlling it  [here](https://github.com/johnfanv2/LenovoLegionLinux/issues/54).
 
-**Other Lenovo Legion models from 2020 to 2023 probably also work. The following models were confirmed. If you have a model with a BIOS version with the same leading letters, e.g. EFCN (like EFCN54WW) then it will probably work. If you want to confirm that your model works or if it does not work, please raise a issue.**
+**Other Lenovo Legion models from 2020 to 2025 probably also work. The following models were confirmed. If you have a model with a BIOS version with the same leading letters, e.g. EFCN (like EFCN54WW) then it will probably work. If you want to confirm that your model works or if it does not work, please raise a issue.**
 
 - Lenovo Legion 5 15IMH05, 15IMH05H (BIOS EFCN54WW): sensors, fan curve, power profile
 - Lenovo Legion 5 15ACH6H (BIOS GKCN58WW or GKCN57WW), Gen 6: sensors, fan curve, power profile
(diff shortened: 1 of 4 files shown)


joppino:
If you set custom mode, choose a preset and then try to write a profile to WMI, the laptop shuts down. I am booting with secure boot

PR #421, opened 2026-04-15
fix: use memory-mapped IO for EC fan curve writes on GKCN65WW

saul-jb:
This PR is a quick tweak from claude which fixed the fan curve writes on Legion 5 Pro 16ACH6H (BIOS GKCN65WW) - sharing here in case it's useful to someone else.

Port IO writes to the fan curve registers are silently ignored on GKCN65WW (and likely other newer BIOS revisions), while reads via port IO continue to work correctly. Switch ec_write_fancurve_legion to use the memory-mapped IO path (ecram_memoryio) which writes directly to EC RAM and works reliably across BIOS versions.

Also fix ecram_memoryio_init being called with physical_ec_start=0 instead of the configured memoryio_physical_ec_start (0xC400). This caused all memory-mapped writes to land at the wrong address, making the path silently broken for all models.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -3083,7 +3083,7 @@ static int ec_read_fancurve_legion(struct ecram *ecram,
 	return 0;
 }
 
-static int ec_write_fancurve_legion(struct ecram *ecram,
+static int ec_write_fancurve_legion(struct ecram_memoryio *ec_memoryio,
 				    const struct model_config *model,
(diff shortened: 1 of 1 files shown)


MrDuartePT:
I can't replicate this on machine with the same bios version.
You also don’t provide step to replicate the problem that LLM thinks it exist on the driver.

I will not accept this or any other PR entirely written by AI, without any proof the problem exist and steps to replicate it.

Issue #422, opened 2026-04-17
Kernel 7.0 driver fails to load


qquique:
I was trying the new kernel 7.0 and this project driver after boot just showed "Loading legion_laptop" and nothing else, when doing a `lsmod | grep -i legion` it was there but no sysfs hwmon, platform_profile or debugfs entries, even blacklisting the new kernel modules lenovo-wmi-gamezone,-other,-helpers,-events it didnt load. 

So i changed what an old todo in code says: 
```c
static const struct acpi_device_id legion_device_ids[] = {
	// todo: change to "VPC2004", and also ACPI paths
	// { "PNP0C09", 0 },
        { "VPC2004", 0 },
	{ "", 0 },
};
```

and it loaded and created the debugfs, hwmon entries, but my question is what we lose/gain here ? as I understand PNP0C09 is direct access, VPC2004 is used by ideapad_laptop driver so I dont know If I have to blacklist also that driver, I tested for some time I didn't see any side effect so far.

and when the legion-laptop driver loaded also created its platform profile in 

```bash
$ ls /sys/class/platform-profile
platform-profile-0  platform-profile-1
```

in platform-profile-0 created by lenovo-wmi-gamezone the choices are : 

```bash
$ cat choices
low-power balanced performance max-power custom
```

in platform-profile-1 created by legion-laptop the choices are : 

```
$ cat choices
quiet balanced balanced-performance performance
```

The equivalence is :
- low-power = quiet
- balanced = balanced
- performance = performance
- custom = balanced-performance
- max-power = #no option#

They are in sync, except by max-power, if you modify in one platform-profile-0 it changes in the other one in platform-profile-1.

Then I wanted to leave the handling of platform-profile to the kernel lenovo-wmi-gamezone and hwmon/fancurves to legion-laptip, but there is a race condition, lenovo-wmi-gamezone doesnt create a platform-profile if legion-laptop loads before it when booting, even when legion-laptop has the flag of not use platform-profiles, I had to add in `/etc/modprobe.d/legion-laptop.conf`

```txt
options legion-laptop enable_platformprofile=0
```

and in `/etc/mkinitcpio.conf`

```txt
MODULES=(i915 lenovo-wmi-gamezone)
```
lenovo-wmi-gamezone starts first and creates its platform-profile and then legion-laptop starts and doesnt create any entry.

And another thing the wmi entries are overlapped between drivers,  I dont know too much of WMI but I think is doing double processing of events? 

```bash
[...]

PR #423, opened 2026-04-26
Adapt to kernel 7.0

qquique:
- Adds powermode Extreme, WMI value 0xE0, EC Value 0x07, exposes it in platform profile.
- Adds flag to model config for handling of extreme powermode.
- Change old powermodes to its equivalence in lenovo-wmi-gamezone (legion.c / legion.py).

```
  lenovo-wmi-gamezone     legion-laptop
  -------------------     -------------
  low-power               quiet
  balanced                balanced
  performance             performance
  custom                  balanced-performance
  max-power               extreme (new)
```
- Change ACPI device id from PNP0C09 to VPC2004, kernel driver in 7.0 acpi-ec binds PNP0C09 and blocks it.
- Change path to look for platform profiles in client legion.py, the old `/sys/firmware/acpi/platform_profile` doesnt accept custom powermode for some reason see `linux/drivers/acpi/platform_profile.c:422`.

Issue related #422

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -6455,8 +6455,11 @@ static SIMPLE_DEV_PM_OPS(legion_pm, NULL, legion_pm_resume);
 // same as ideapad
 static const struct acpi_device_id legion_device_ids[] = {
 	// todo: change to "VPC2004", and also ACPI paths
-	//{ "PNP0C09", 0 },
-	{ "VPC2004", 0 },
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(7, 0, 0)
+	    { "VPC2004", 0 },
+#else
+	    { "PNP0C09", 0 },
+#endif
 	{ "", 0 },
 };
 MODULE_DEVICE_TABLE(acpi, legion_device_ids);
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -210,6 +210,7 @@ struct model_config {
 	// TODO: maybe use bitfield
 	bool has_minifancurve;
 	bool has_custom_powermode;
+	bool has_extreme_powermode;
 	enum access_method access_method_powermode;
 
 	enum access_method access_method_keyboard;
@@ -955,6 +956,7 @@ static const struct model_config model_lzcn = {
 	.memoryio_size = 0x300,
 	.has_minifancurve = true,
 	.has_custom_powermode = true,
+	.has_extreme_powermode = true,
 	.access_method_powermode = ACCESS_METHOD_WMI,
 	.access_method_keyboard = ACCESS_METHOD_WMI2,
 	.access_method_fanspeed = ACCESS_METHOD_WMI3,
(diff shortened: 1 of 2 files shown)


chinese-wzq:
Hi, Happy Labour Day! I've also been working on VPC2004 support, and my changes involve a larger driver rework (see PR #427 ). Since my model also supports Extreme mode, and I'm on kernel 7.0 now, I was thinking we could integrate the Extreme mode and other new features into my PR. I'd be happy to handle the rebase and integration, and of course I'll make sure you're properly credited (e.g., Co-authored-by). If you're up for it, your review would be really valuable. What do you think?

kernel_module/legion-laptop.c
```diff
@@ -6419,7 +6449,8 @@ static SIMPLE_DEV_PM_OPS(legion_pm, NULL, legion_pm_resume);
 // same as ideapad
 static const struct acpi_device_id legion_device_ids[] = {
 	// todo: change to "VPC2004", and also ACPI paths
-	{ "PNP0C09", 0 },
+	//{ "PNP0C09", 0 },
+	{ "VPC2004", 0 },
```
MrDuartePT: This change complete breaks the driver on my machine (including fan curve), maybe is because of the older kernel is better to add this fix:
```c
static const struct acpi_device_id legion_device_ids[] = {
	// todo: change to "VPC2004", and also ACPI paths
	#if LINUX_VERSION_CODE >= KERNEL_VERSION(7, 0, 0)
	    { "VPC2004", 0 },
    #else
	    { "PNP0C09", 0 },
    #endif
	{ "", 0 },
};
```
Do a similar thing for the python side.
```python
kernel_version = tuple(map(int,os.uname().release.split('-')[0].split('.')))

DEFAULT_ENCODING = "utf8"
DEFAULT_CONFIG_DIR = "/etc/legion_linux"
if kernel_version >= (7, 0, 0):
    LEGION_SYS_BASEPATH = '/sys/module/legion_laptop/drivers/platform:legion/VPC2004:00'
else:
    LEGION_SYS_BASEPATH = '/sys/module/legion_laptop/drivers/platform:legion/PNP0C09:00'
IDEAPAD_SYS_BASEPATH = '/sys/bus/platform/drivers/ideapad_acpi/VPC2004:00'
LBLDVC_FILE = "/sys/firmware/efi/efivars/LBLDVC-871455d1-5576-4fb8-9865-af0824463c9f"
LBLDESP_FILE = "/sys/firmware/efi/efivars/LBLDESP-871455d0-5576-4fb8-9865-af0824463b9e"
```

Aside for that everting else works fine. The only thing I can't test is the extreme mode (2021 legion).

qquique: done

chinese-wzq:
Hi, just saw your PR is already under review — great!
I'll hold off on my side and simply rebase onto it once yours gets merged. That way we avoid unnecessary conflicts. I'll test the Extreme mode from your PR later as well. Looking forward to your changes landing. Cheers!

chinese-wzq:
After testing on my model with the flag enabled, the CPU power draw remained nearly unchanged during a stress test, but the GPU consistently pulled an extra 15W on average – that confirms Extreme mode is working. Good job! By the way, would it be possible to also add the `has_extreme_powermode` flag for `model_lpcn`, which is my machine?

qquique:
Changes
- For kernel >= 7.0 driver loads with a virtual device instead of linking to an acpi device, avoiding the ACPI_COMPANION.
- For kernel < 7.0 keeps using the PNP0C09 device.
- Moved static ACPI paths to a defaults array and helper enum of ACPI paths used in the driver. Works for any kernel
- Modified config_model to add an array of ACPI paths to override the defaults, so the paths can be configured by model, example model_lzcn in code. Works for any kernel.
- Created a helper function to select the ACPI path from the override, if not found uses the default.
- Changed acpi function header parameter from acpi_handle to acpi_device to handle The Handle :) in 2 functions instead of coming from upper levels.
- updated path for the virtual device in legion_gui.py for kernel 7.0.

MrDuartePT:
Looks good!
Closes: #422

Eoligan:
Hi! Just a heads-up for anyone using the C daemon (legiond) on Kernel 7.0.

I noticed that even with the latest updates, the fan curves don't trigger in Quiet mode. After some debugging, it seems legiond is still looking for the string "quiet", but the kernel is now reporting it as "low-power" (as mentioned in the equivalence table above).

I'm no expert, but I managed to fix it by adding a match for "low-power" in extra/service/legiond/modules/powerstate.c (around line 44):

if (MATCH(profile, "quiet") || MATCH(profile, "low-power"))

After recompiling, everything works perfectly. Just sharing in case it helps

MrDuartePT:
Yes that is true. I complete forget to updated legiond, thanks for noticing.

PR #425, opened 2026-04-29
Fix kernel NULL pointer dereference when enable_platformprofile is false

qquique:
When driver is loaded with `option enable_platformprofile=0` and user changes power profile / Fn+Q, function `legion_platform_profile_notify` generates a kernel BUG error.

Happens with acpi_device_ids PNP0C09 or VC2004, tested on Kernel LTS 6.18 and 7.0.x.

```
kernel: BUG: kernel NULL pointer dereference, address: 0000000000000030
kernel: #PF: supervisor read access in kernel mode
kernel: #PF: error_code(0x0000) - not-present page
```

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -4971,8 +4971,10 @@ static void legion_platform_profile_notify(struct device *dev)
 static void legion_platform_profile_notify(void)
 #endif
 {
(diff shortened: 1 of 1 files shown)


MrDuartePT:
Thanks for the fix!
I never tested that option to be fair.

PR #431, opened 2026-05-04
Add max_rpm field to model_config and fancurve structures

chinese-wzq:
- Introduced a new max_rpm field in the model_config structure to specify the maximum fan speed in RPM.
- Updated the fancurve structure to include max_rpm for PWM conversion.
- Modified fan speed calculation functions to utilize the max_rpm field, allowing for custom maximum RPM settings.
- Adjusted debugfs output and fan curve reading functions to reflect the new max_rpm functionality.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -223,6 +223,10 @@ struct model_config {
 
 	phys_addr_t ramio_physical_start;
 	size_t ramio_size;
+
+	// Maximum fan speed in RPM, used for fancurve PWM conversion.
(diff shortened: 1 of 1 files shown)


st0nie:
Thanks for working on this. A per-model maximum RPM makes sense and would be useful.

The PR now conflicts with `main`, so could you rebase it and document where the 5400 RPM value came from? `model_lpcn` is shared by LPCN and M3CN DMI entries, so the value should either be confirmed for both or the model configurations should be split.

After rebasing, current CI plus LPCN/M3CN hardware readback for `fan*_max` and PWM/RPM conversion would help confirm it.

PR #443, opened 2026-05-11
fan_unlock: lift firmware fan ceiling via WMAA(0,0x0D) — closes the 4400 RPM cap on KWCN54WW (#429)

AndresIslas99:
## Summary

Adds a new `fan_unlock` sysfs (and matching `legion_cli fan-unlock-{enable,disable,status}` Python wrappers) that issues `WMAA(0, 0x0D, arg)` on the GameZone WMI GUID. On firmwares like **KWCN54WW** (Legion Pro 7 16IRX8H), this lifts the firmware-imposed fan ceiling that has been capping fans at **~4400 RPM** under Linux even in Performance mode, raising the ceiling to **~7100 RPM** — the EC's high-end fan curve subtable.

This closes a long-standing question raised in #429, which I had previously concluded (incorrectly) ended at \"absolute anti-tamper, requires DXE-time IHISI handshake\". The actual unlock turned out to be one WMI sub-command that wasn't enumerated. **The DXE/IHISI/Vantage-handshake hypothesis was wrong** — full retraction posted on the issue with measurements.

## What changes

| File | Change |
|---|---|
| `kernel_module/legion-laptop.c` | New `WMI_METHOD_ID_FAN_EXTREME_TOGGLE = 13` constant; new `fan_unlock_show/store` + `DEVICE_ATTR_RW(fan_unlock)`; registered in `legion_sysfs_attributes[]`. |
| `python/legion_linux/legion_linux/legion.py` | New `FanUnlock(BoolFileFeature)` mirroring the `LockFanController` pattern; wired into `LegionModelFacade`. |
| `python/legion_linux/legion_linux/legion_cli.py` | New `FanUnlockFeatureCommand` with `enable/disable/status` subcommands; registered in `main()`. |
| `README.md` | One-line entry under Features. |

Total diff: **+85 lines**, no removals, no refactors of existing code paths.

## How the unlock works

Disassembly of `WMAA` (DSDT line 36638 on KWCN54WW) showed a sub-command at `Arg1=0x0D` that toggles `^^PC00.LPCB.EC0.NCMD(0x59, 0x77 if Arg2==1 else 0x76)`. `NCMD` writes the command byte to **port 0x6C** and the data byte to port 0x68 — the **alt-namespace EC interface** the standard driver doesn't touch. Inside the EC, `0x59 0x77` raises the fan curve subtable index from the locked low subtable (PWM ~33%, ~4400 RPM) into the high subtables (PWM ~54%, ~7000 RPM) that prior investigation identified as \"EXTREME curves\" only reachable from firmware.

Two related sub-commands also raise fans but are less stable:
- `Arg1=0x16, Arg2=1` → `NCMD(0x59, 0x78)` → peaks ~6700, decays to ~6000 under sustained load
- `Arg1=0x19, Arg2=1` → `NCMD(0x59, 0x7B)` → peaks ~5600, decays to ~4900

Only `0x0D, 1` → `NCMD(0x59, 0x77)` holds steady at 7100 RPM. **Exposing only this one as the user-facing knob.**

## Validation (Legion Pro 7 16IRX8H, BIOS KWCN54WW, EC 0x5507, kernel 6.8.0-110-generic)

```
$ echo 1 | sudo tee /sys/devices/pci0000:00/0000:00:1f.0/PNP0C09:00/fan_unlock
$ cat /sys/class/hwmon/hwmon3/fan1_input /sys/class/hwmon/hwmon3/fan2_input
7000
7100

$ sudo legion_cli fan-unlock-status
True
$ sudo legion_cli fan-unlock-disable
$ cat /sys/class/hwmon/hwmon3/fan1_input
3500   # back to firmware default cap
```

`stress-ng --cpu 24` for 60s with the unlock active:

| Metric | Cap (4400 RPM) | Unlocked (7100 RPM) |
|---|---|---|
| Peak package temp | 86°C @ PL1=35W | **81°C @ PL1=40W** |
| Stable temp under load | 76-77°C | **72°C** |
| Sustained max freq | 3.8GHz (thermal-throttled) | **5.0GHz full** |
| Throttle states triggered | warn | **none** |

Read-back returns the last-applied value; invalid input is rejected with `-EINVAL` by `kstrtobool`, last good value preserved. WMI failures surface as `-EIO` with no side effects on fans.

## Compatibility

- The WMAA sub-command exists on KWCN's BMOF (`SETSMARTFANMODE` at id 44 / `0x2C`, FAN_EXTREME_TOGGLE at id 13 / `0x0D`). Other firmwares without this sub-command will return an ACPI error and the write surfaces as `-EIO` — fan state unchanged. **No risk to firmwares that don't support it**, the user just gets an EIO and learns it's not supported on their model.
- Tested only on KWCN54WW (16IRX8H). Owners of other Legion 2022/2023 BMOFs that include id 13 in the GameZone GUID should be able to confirm. Happy to add a `model_config` flag (`has_fan_unlock`) and gate sysfs creation on it if you'd prefer that over surfacing on all models.

## Note on building `main` against current Linux

While preparing this branch I found that unmodified `main` does not compile on kernel 6.8 because line ~5414 references `PLATFORM_PROFILE_MAX_POWER` from inside a `LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)` block, but the symbol only exists from 6.19+. **Unrelated to this PR** (the code path was untouched). Functional verification of this PR was performed against a tree carrying a local one-line fix for that guard mismatch. Happy to file it separately as a one-character nudge of the `#if` to `>= 6, 19, 0` if it'd help.

## References

- Original issue / multi-comment investigation history: #429
- WMAA sub-command map: DSDT lines 36638-37800 on KWCN54WW
- EC fan curve subtables (firmware bank 1, 0x1E000-0x1FE60): subtables 0/7/14 are the locked low (PWM 0x21-0x23 = 4400 RPM), subtables 8-11/15-17 are the high (PWM 0x36 = ~7000 RPM); the unlock just flips the index.

Happy to iterate on naming (`fan_unlock` vs `fan_extreme` vs `fan_high_subtable`), gating, or any other aspect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ It allows you to control features like the fan curve, power mode, power limits, 
   - Set acceleration and deceleration for each the fan when the fan speed should increase or decrease
   - Save and load presets for different modes
 - [X] Lock and unlock the fan controller and fan speed
+- [X] Lift the firmware-imposed fan ceiling on supported models (`fan_unlock` sysfs / `legion_cli fan-unlock-{enable,disable,status}`). On the Legion Pro 7 16IRX8H (BIOS KWCN54WW) this raises the cap from ~4400 RPM to ~7100 RPM. Discovered via `WMAA(0, 0x0D, 0x01)` — see issue #429.
 - [X] Switch power mode (quiet, balanced, performance) using software
   - Now you can do it using software in your system settings
   - Changing with `Fn+Q` is also possible
--- a/extra/scripts/wmaa-explore.sh
+++ b/extra/scripts/wmaa-explore.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+# WMAA sub-command explorer for Lenovo Legion firmwares.
+# Iterates GameZone WMI WMAA(0, mid, val) sub-commands, captures fan response,
+# and produces a compatibility report. Use this on un-tested firmwares to find
+# fan-unlock-equivalent commands.
+#
+# DISCOVERED SO FAR (see issue #429 / PR #443):
+#   KWCN54WW (Legion Pro 7 16IRX8H):
+#     WMAA(0, 0x0D, 1) → fans 4400 → 7100 RPM (unlock — exposed as fan_unlock)
+#     WMAA(0, 0x16, 1) → fans 4400 → 6700 RPM (decays to ~6000)
+#     WMAA(0, 0x19, 1) → fans 4400 → 5600 RPM (decays to ~4900)
+#
+# Goal: build a per-firmware compat matrix to cover all Legion 2022/2023
+# models in the upstream driver.
+#
+# Safety:
+#   - Skips known-destructive sub-commands (0xE0, 0xFF — see README in PR #443
+#     comments for context).
+#   - Each test reverses with WMAA(0, mid, 0) before the next.
+#   - Aborts immediately if fans drop below baseline (anti-tamper triggered).
+
+set -u
+
+GAMEZONE=/sys/devices/pci0000:00/0000:00:1f.0/PNP0C09:00/gamezone_method
+HWMON=
+for d in /sys/class/hwmon/hwmon*/; do
+    [ "$(cat "${d}name" 2>/dev/null)" = "legion_hwmon" ] && HWMON="${d%/}" && break
+done
+
+if [ -z "$HWMON" ] || [ ! -w "$GAMEZONE" ]; then
+    echo "ERROR: legion-laptop driver not loaded or not patched. Need gamezone_method sysfs."
+    echo "Apply patches from https://github.com/johnfanv2/LenovoLegionLinux issue #429"
+    exit 1
+fi
+
+readonly REPORT=/tmp/wmaa-compat-report.txt
+> "$REPORT"
+
+echo "=== WMAA explorer for $(cat /sys/class/dmi/id/product_name) BIOS=$(cat /sys/class/dmi/id/bios_version) ==="
+echo "Logging to $REPORT"
+echo
+
+read_fans() { echo "$(cat $HWMON/fan1_input)/$(cat $HWMON/fan2_input)"; }
+read_cpu() { awk '/^Package id 0:/{gsub(/[+°C]/,"",$4);print int($4);exit}' < <(sensors coretemp-isa-0000 2>/dev/null); }
+
+baseline_fans=$(read_fans)
+echo "Baseline (idle): fans=$baseline_fans cpu=$(read_cpu)°C" | tee -a "$REPORT"
+echo
+
+# Generate light load so EC actually wants to spin fans
+stress-ng --cpu 16 --timeout 240s > /dev/null 2>&1 &
+STRESS=$!
+trap "kill $STRESS 2>/dev/null; echo 0x0D:0x00 > $GAMEZONE 2>/dev/null" EXIT INT TERM
+sleep 8
+load_fans=$(read_fans)
+echo "Under light load (16 CPUs): fans=$load_fans cpu=$(read_cpu)°C" | tee -a "$REPORT"
+echo
+
+# DANGEROUS sub-commands to skip — known to either crash, drop fans, or
+# require a multi-step protocol that we shouldn't trigger blind.
+DANGER_LIST="0xE0 0xFF 0x8C 0xCD"
+is_dangerous() {
+    local v=$1
+    for d in $DANGER_LIST; do [ "$v" = "$d" ] && return 0; done
+    return 1
+}
+
+echo "Testing WMAA(0, mid, 1) for mid in 0x00..0x40, then back to 0..." | tee -a "$REPORT"
+echo "(skipping known-destructive: $DANGER_LIST)" | tee -a "$REPORT"
+echo
+
+for n in $(seq 0 64); do hex=$(printf "0x%02X" $n)
+    if is_dangerous "$hex"; then
+        printf "  %s  SKIPPED (destructive)\n" "$hex" | tee -a "$REPORT"
+        continue
+    fi
+    echo "$hex:0x01" > "$GAMEZONE" 2>/dev/null
+    sleep 3
+    fans=$(read_fans)
+    cpu=$(read_cpu)
+    f1=${fans%/*}
+    base1=${load_fans%/*}
+    delta=$((f1 - base1))
+    flag=""
+    [ $delta -gt 500 ] && flag="🚀 RAISES FANS"
+    [ $delta -lt -500 ] && { flag="⚠ DROPS FANS"; }
+    printf "  %s  fans=%s  cpu=%s°C  delta=%+d  %s\n" "$hex" "$fans" "$cpu" "$delta" "$flag" | tee -a "$REPORT"
+    # Always disable before next to avoid stacking effects
+    echo "$hex:0x00" > "$GAMEZONE" 2>/dev/null
+    sleep 1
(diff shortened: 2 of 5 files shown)


AndresIslas99:
## Why this PR exists / context for #429 reviewers

Cross-referencing because issue #429 (which this PR closes the question for) is locked and I can't post the retraction there.

In #429 I previously wrote 12 progressively-more-confident comments concluding the fan cap was an absolute firmware anti-tamper requiring a DXE-time IHISI handshake. **That was wrong.** The unlock turned out to be a single WMI sub-command (`WMAA(0, 0x0D, 0x01)` on the GameZone GUID) that I had not enumerated. It maps to `^^PC00.LPCB.EC0.NCMD(0x59, 0x77)` on the EC's alt-namespace command port (0x6C/0x68) — exactly the path I correctly identified as Vantage's, but reachable through a plain WMI call rather than the SMM machinery I was chasing.

### What unlocks the fans

```
echo "0x0D:0x01" | sudo tee /sys/.../gamezone_method   # via the existing probe sysfs
# or, with this PR:
echo 1 | sudo tee /sys/.../fan_unlock
sudo legion_cli fan-unlock-enable
```

### Measurements on KWCN54WW (Legion Pro 7 16IRX8H)

| | Cap (4400 RPM) | Unlocked (7100 RPM) |
|---|---|---|
| Fan ceiling under load | 4400 RPM | **7100 RPM** (+62%) |
| `stress-ng --cpu 24` 60s peak | 86°C @ PL1=35W | **81°C @ PL1=40W** |
| Stable temp under sustained load | 76-77°C | **72°C** |
| Sustained max freq | 3.8 GHz (throttled) | **5.0 GHz full** |

`stress-ng --cpu 24 --timeout 60s` log with unlock active, package temp + both fans + freq:

```
t+5s   cpu=79°C  fan=7000/7100  max=5000MHz
t+15s  cpu=81°C  fan=7000/7100  max=5000MHz
t+25s  cpu=73°C  fan=7000/7100  max=5000MHz
t+45s  cpu=72°C  fan=7000/7100  max=5000MHz
t+60s  cpu=72°C  fan=7000/7100  max=5000MHz
```

### Compatibility request

I only have one machine to test (`82WQ` / KWCN54WW). Other Legion 2022/2023 model owners — if you check out this branch (or just use the probe sysfs already present in the repo), please report back here with:

- Your model + DMI + BIOS string
- Baseline `cat /sys/class/hwmon/hwmon*/fan1_input` under `stress-ng --cpu 16` for ~10s
- After `echo 1 > /sys/.../fan_unlock`, same reading

I'd like to build a compatibility matrix before merge so the maintainers can decide whether to gate the sysfs on a `model_config.has_fan_unlock` flag or expose it everywhere with an EIO fallback. Happy to add the gate either way.

### Apology

Closing #429 with a wrong "absolute" conclusion was net-negative for the community — anyone searching the issue tracker for this problem would have read the conclusion and given up. Reopening was blocked (issue is locked), hence this comment as a beacon. The investigation infrastructure (BIOS extraction, EC firmware analysis, IHISI kernel module, ITE register cross-reference from System76) was real and may still be useful for other firmware questions, but on the specific question of "can we lift the 4400 RPM cap on Linux" — yes, one WMI call.

AndresIslas99:
## Follow-up: 6-level granular fan speed scale discovered (WMAA 0x0D–0x12)

While building a community-facing sub-command enumerator (`extra/scripts/wmaa-explore.sh`, just pushed to this branch), the explorer found that **the WMAA fan-control range is not a single toggle but a 6-level scale** sitting at sub-commands 0x0D through 0x12, all with arg=1.

Methodology: under steady `stress-ng --cpu 16` load, applied each sub-command, waited 4s for fans to stabilize, recorded RPM, then disabled before the next test. Results on KWCN54WW (Legion Pro 7 16IRX8H):

| sub-cmd | NCMD code (per DSDT) | Steady fans (RPM) | vs default cap (4400) |
|---|---|---|---|
| **0x0D=1** | NCMD(0x59, 0x77) | **7000–7100** | **+62%** (max EXTREME) |
| 0x0E=1 | (probably NCMD 0x59, 0x7C) | 6400–6500 | +47% |
| 0x0F=1 | (probably NCMD 0x59, 0x7D) | 5900–6000 | +36% |
| 0x10=1 | (probably NCMD 0x59, 0x7E) | 5200–5400 | +20% |
| 0x11=1 | (probably NCMD 0x59, 0x7F) | 4600–4800 | +9% |
| 0x12=1 | (probably NCMD 0x59, 0x80) | 4000–4100 | ~at cap |
| 0x13=1 | (probably NCMD 0x59, 0x81) | 3500–3600 | **−18%** (below cap) |

Note: my DSDT only directly disasms NCMD codes 0x76–0x7B (toggles for sub 0x0D, 0x16, 0x19) — the 0x7C+ NCMD codes for sub-commands 0x0E–0x13 are inferred by adjacency; would need a fresh DSDT walk to confirm. Either way, the **observed empirical effect is a clean monotonic 6-step scale**.

This means the eventual upstream API can be richer than the boolean `fan_unlock` this PR exposes — could be a `fan_speed_level` int 0–5 (or similar) that gives users granular control without entering Custom mode (which still triggers the anti-tamper documented earlier in #429).

### Why not change this PR

I'm proposing we **keep this PR as the boolean-only `fan_unlock` for landing**, because:

1. The boolean is what 95% of users actually want — \"give me max fans, my pasta is degraded\" — and matches the existing `lockfancontroller` pattern.
2. The granular scale needs cross-model verification before exposing it as an API. On other firmwares the levels may map differently or not exist.
3. A separate follow-up PR can add `fan_speed_level` once the explorer has reports from 2–3 different KWCN-family firmwares.

The explorer script (`extra/scripts/wmaa-explore.sh`) is what makes that follow-up possible — anyone with a Legion can run it and post their report here. I've added it to this PR because it's both how the discovery was made and the tool the community needs to extend it.

### Stability of `fan_unlock` re-validated

After the explorer's 240+ second stress run with 20 different sub-commands toggling on/off, `fan_unlock=1` still holds steady at 7000–7100 RPM. No driver oops, no kernel WARN, no anti-tamper trigger. Logs clean.

```bash
$ sudo cat /sys/.../fan_unlock
1
$ cat /sys/class/hwmon/hwmon3/fan1_input /sys/class/hwmon/hwmon3/fan2_input
7000
7100
$ dmesg | grep -i legion-laptop | tail -5
# (no warnings or errors)
```

st0nie:
Review: useful feature (closes #429), but not merging as-is: fan_unlock is exposed unconditionally in legion_sysfs_attributes for every model, so WMAA(0, 0x0D, x) can be invoked on firmwares where sub-command 0x0D is unverified (your own explorer marks several nearby sub-commands as destructive). Please gate the attribute behind a model/BIOS allowlist (e.g. a model_config flag enabled only for validated firmwares like KWCN) so unverified machines don't get the knob.

st0nie:
Thanks for all the investigation here. The fan unlock itself is useful, and the KWCN54WW test results look promising.

I still don’t feel comfortable exposing `WMAA(0, 0x0D, x)` on every supported model when it has only been validated on KWCN54WW. Could you add a model/BIOS capability flag and hide the sysfs attribute everywhere else?

I’d also prefer to leave `wmaa-explore.sh` out of this PR. It sends unknown commands `0x00..0x40` on untested firmware, assumes `0` reverses each command, and the interrupt trap only resets `0x0D` instead of the currently active command.

A focused PR with the gated kernel feature, Python/CLI wrapper, and normal CI would be much safer and easier to merge.

PR #450, opened 2026-05-27
platform_profile: expose balanced-performance as alias for custom

thewraith420:
## Summary

Commit 271707f renamed the fourth Legion power mode from `PLATFORM_PROFILE_BALANCED_PERFORMANCE` to `PLATFORM_PROFILE_CUSTOM` to align with `lenovo-wmi-gamezone`'s naming scheme. While this matches Lenovo's official tooling, it silently broke a set of pre-existing userspace tools and configurations that referenced the kernel's standard `balanced-performance` name — most notably:

- **TLP** — its documentation in `/usr/share/tlp/defaults.conf` lists `balanced-performance` as an example value and many existing `/etc/tlp.conf` files were configured with `PLATFORM_PROFILE_ON_AC=balanced-performance`. Those configs now silently fail on the new driver.
- **power-profiles-daemon** and GNOME UI flows that target the standard kernel profile names.
- User scripts, autostart entries, and muscle memory.

This PR exposes **both** names simultaneously, so the new naming alignment with `lenovo-wmi-gamezone` is preserved and backward-compatible userspace tooling keeps working.

## Changes

Purely additive — no behavior change for existing callers using `custom`.

- **`choices`** (both new probe-callback API and legacy API): set bits for both `PLATFORM_PROFILE_CUSTOM` *and* `PLATFORM_PROFILE_BALANCED_PERFORMANCE`. Both names appear in `/sys/class/platform-profile/platform-profile-0/choices` and both are writable.
- **`legion_platform_profile_set()`**: accepts both enums via fallthrough; both map to `LEGION_WMI_POWERMODE_CUSTOM`. Hardware state is identical regardless of which name was used.
- **`legion_platform_profile_get()`**: unchanged. Reads always return `PLATFORM_PROFILE_CUSTOM`, so `lenovo-wmi-gamezone` and Lenovo Vantage see the canonical name.

## Test plan

Tested on Lenovo LOQ 15IRH8 (LZCN BIOS), Linux 7.0.9 with LenovoLegionLinux master HEAD + this patch applied:

- [x] `cat /sys/class/platform-profile/platform-profile-0/choices` shows both `balanced-performance` and `custom` in the list
- [x] `echo custom > .../profile` works; subsequent read returns `custom`
- [x] `echo balanced-performance > .../profile` works; subsequent read returns `custom` (canonical)
- [x] TLP with `PLATFORM_PROFILE_ON_AC=balanced-performance` correctly switches the LED indicator on plug-in
- [x] AC plug/unplug cycle still drives mode switching via the legion driver's virtual platform device

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -5338,6 +5338,11 @@ static int legion_platform_profile_set(struct platform_profile_handler *pprof,
 		powermode = LEGION_WMI_POWERMODE_LOW_POWER;
 		break;
 	case PLATFORM_PROFILE_CUSTOM:
+	case PLATFORM_PROFILE_BALANCED_PERFORMANCE:
+		/* Accept both names: "custom" (matches lenovo-wmi-gamezone) and
+		 * "balanced-performance" (the kernel's standard naming used by
+		 * TLP, power-profiles-daemon, and pre-existing user scripts).
+		 */
 		powermode = LEGION_WMI_POWERMODE_CUSTOM;
 		break;
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 19, 0)
@@ -5362,8 +5367,12 @@ static int legion_platform_profile_probe(void *drvdata, unsigned long *choices)
 	set_bit(PLATFORM_PROFILE_LOW_POWER, choices);
 	set_bit(PLATFORM_PROFILE_BALANCED, choices);
 	set_bit(PLATFORM_PROFILE_PERFORMANCE, choices);
-	if (conf_has_custom_powermode && conf_access_method_powermode == ACCESS_METHOD_WMI)
+	if (conf_has_custom_powermode && conf_access_method_powermode == ACCESS_METHOD_WMI) {
 		set_bit(PLATFORM_PROFILE_CUSTOM, choices);
+		/* Expose "balanced-performance" as an alias so userspace tools
+		 * that pre-date the rename to "custom" keep working unchanged. */
+		set_bit(PLATFORM_PROFILE_BALANCED_PERFORMANCE, choices);
+	}
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 19, 0)
 	if (conf_has_extreme_powermode && conf_access_method_powermode == ACCESS_METHOD_WMI)
 		set_bit(PLATFORM_PROFILE_MAX_POWER, choices);
@@ -5408,6 +5417,10 @@ static int legion_platform_profile_init(struct legion_private *priv)
 	if (priv->conf->has_custom_powermode &&
 	    priv->conf->access_method_powermode == ACCESS_METHOD_WMI) {
 		set_bit(PLATFORM_PROFILE_CUSTOM,
+			priv->platform_profile_handler.choices);
+		/* Expose "balanced-performance" as an alias so userspace tools
+		 * that pre-date the rename to "custom" keep working unchanged. */
+		set_bit(PLATFORM_PROFILE_BALANCED_PERFORMANCE,
 			priv->platform_profile_handler.choices);
 	}
 	if (priv->conf->has_extreme_powermode &&


thewraith420:
Thanks for the review — both comments raise valid points to discuss. Responding to each:

**Re: `PLATFORM_PROFILE_BALANCED_PERFORMANCE` requiring a 6.9 guard**

The bot is mistaken about the introduction kernel. Per `Documentation/ABI/testing/sysfs-platform_profile` in mainline:

> `What: /sys/firmware/acpi/platform_profile_choices`
> `Date: October 2020`
> [...]
> `balanced-performance    Balance between performance and low power consumption with a slight bias towards performance`

`PLATFORM_PROFILE_BALANCED_PERFORMANCE` has been in the kernel since v5.10 (October 2020). That's well before any kernel this driver currently supports — the lowest `LINUX_VERSION_CODE` check in `legion-laptop.c` today is `6.13`. So no guard is needed; the enum is always available.

Notably, before commit 271707f this file used `PLATFORM_PROFILE_BALANCED_PERFORMANCE` directly without any version guard, and the driver built fine on every supported kernel.

**Re: round-trip mismatch / dual exposure**

The concern is real but I'd argue the platform_profile API does not actually guarantee read-after-write identity:

- `Documentation/userspace-api/sysfs-platform_profile.rst` states: *"Drivers must map their internal profile representation onto this fixed set."* Mapping is explicitly expected — input strings can resolve to a different output state.
- The kernel core itself normalizes: when multiple platform_profile handlers disagree, the aggregate `/sys/firmware/acpi/platform_profile` emits the literal string `custom` to signal the conflict — already a case where what you read doesn't match what any single handler was written.
- Other drivers expose aliases too. For example, several thermal drivers accept both legacy and current profile names and converge them internally.

In practice, `power-profiles-daemon` does not perform read-after-write verification — it writes the configured profile string and trusts the kernel to do the right thing. The same is true for TLP. The only consumer that does verify is `lenovo-wmi-gamezone` and Lenovo Vantage, which both expect `custom` — and those continue to work unchanged because the getter is intentionally left alone to return the canonical name.

That said, I'm happy to gate the alias behind a module parameter (e.g., `legacy_balanced_performance_alias=N` default off) if you'd prefer the opt-in model. Let me know which direction you'd like and I'll update the PR.

qquique:
To consider : 
- https://lore.kernel.org/platform-driver-x86/20251026081240.997038-3-derekjohn.clark@gmail.com/
- https://github.com/torvalds/linux/commit/8d0aef2b3a17e5ed3afc8c9a2e71f22ac41a799f

- balanced-performance in previous kernel versions without this driver 'LLL' was setting in some cases the platform profile to performance.
- tlp and power-profiles daemon reads from `/sys/firmware/acpi/platform_profile_choices`, and writes to `/sys/firmware/acpi/platform_profile`. `platform_profile` [doesnt accept](https://github.com/torvalds/linux/blob/master/drivers/acpi/platform_profile.c#L422-L423) the value 'custom'.
- With this driver 'LLL' and lenovo-wmi-gamezone reads from /sys/class/platform-profile/platform-profile-?/choices and writes to /sys/class/platform-profile/platform-profile-?/profile, profile accepts the value 'custom' 

So:
- tlp.conf says :

> ```txt
> # check the output of tlp-stat -p to determine availability on your
> # hardware and additional profiles such as: balanced-performance, quiet, cool 
> ```

The choices for the user: 
- balanced-performance, without this driver on previous kernels on some cases was set to performance.
- balanced-performance, with this driver should be set to custom, and because it tries to write `custom` on `/sys/firmware/acpi/platform_profile` nothing will happen. 

I personally use tlp and just have it configured  to balanced-performance (does nothing), but the driver LLL I always leave it on custom once and it stays between reboots/shutdowns on different kernel versions.

power-profiles-daemon: 
reads for what acpi provides and groups them in three values : quiet / balanced / performance ([see](https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/blob/main/src/ppd-driver-platform-profile.c#L58-L103)) and writes based on those three values , same as what Fn+Q does, so here balanced-performance doesnt matter.

thewraith420:
Thanks @qquique — the link to Derek's `8d0aef2` is helpful context, and confirming that the aggregate `/sys/firmware/acpi/platform_profile` explicitly returns `-EINVAL` on `custom` actually reinforces why this PR matters in practice. Quick clarification on the use case:

Your "set custom once and let it persist" workflow works perfectly when a single platform_profile fits the whole session. The case this PR addresses is different: **LLL doesn't auto-switch the platform_profile between AC and battery power**, so users rely on TLP (or power-profiles-daemon) to handle that transition.

That auto-switching happens through the aggregate sysfs — which, as you pointed out, rejects `custom` at line 422-423:

```c
i = sysfs_match_string(profile_names, buf);
if (i < 0 || i == PLATFORM_PROFILE_CUSTOM)
    return -EINVAL;
```

So a TLP config of:

```
PLATFORM_PROFILE_ON_AC=custom
PLATFORM_PROFILE_ON_BAT=low-power
```

silently fails on AC — the aggregate rejects `custom` outright, and TLP doesn't have a path to the per-driver sysfs. The user is stuck either:

- Picking one profile and never switching (your workflow), or
- Configuring TLP with a name the aggregate accepts (which currently doesn't include any name that maps to `LEGION_WMI_POWERMODE_CUSTOM`).

With this PR's alias:

```
PLATFORM_PROFILE_ON_AC=balanced-performance  # → LEGION_WMI_POWERMODE_CUSTOM
PLATFORM_PROFILE_ON_BAT=low-power
```

works end-to-end through the aggregate. TLP writes `balanced-performance`, the aggregate accepts it (since the alias puts it in the choice list), LLL maps it to `LEGION_WMI_POWERMODE_CUSTOM`, hardware ends up in the right state on each power transition. This matches what TLP's own `defaults.conf` documents:

```
# Note: check the output of tlp-stat -p to determine availability on your
# hardware and additional profiles such as: balanced-performance, quiet, cool.
```

The LED-color UX is a nice secondary benefit (and aligns with the motivation in `8d0aef2`), but the primary problem is that without the alias there is no way to get LLL's `custom` hardware state through any AC/battery-aware userspace tool that writes to the aggregate sysfs.

Let me know which way you'd like to go — keep the choices alias as-is, or gate it behind a module parameter as Copilot suggested. Either works for me.

qquique:
> PLATFORM_PROFILE_ON_BAT=low-power

With TLP when I unplug the AC the laptop goes to quiet/low-power, that is triggered by the bios (blue led), in `/sys/firmware/acpi/platform_profile` stays as custom , and when I plug again the machine goes back to custom mode (purple led).

to expand a bit : 
```bash
# this returns error
$ echo custom > /sys/firmware/acpi/platform_profile
# so with LLL, change to custom
$ echo custom > /sys/class/platform-profile/platform-profile-0/profile
# notifies to acpi
$ cat /sys/firmware/acpi/platform_profile
custom
```

thewraith420:
After your reply I ran a clean test that fully isolates BIOS behavior from userspace on my LOQ 15IRH8:

- `systemctl stop tlp`
- Moved `/usr/lib/udev/rules.d/85-tlp.rules` aside and `udevadm control --reload-rules`
- Set `TLP_ENABLE=0` in `tlp.conf` for good measure, then rebooted

With absolutely nothing in userspace touching `platform_profile`, plugging/unplugging AC still drives the LED color correctly via BIOS/EC firmware. Kernel `/sys/firmware/acpi/platform_profile` stays at `custom` (as you described on your hardware), but the visible behavior — the LED following power state — works flawlessly without any kernel-side intervention.

So my primary argument here ("LLL doesn't auto-switch, TLP needs the alias to handle AC/BAT") is empirically wrong for at least this hardware. The BIOS does handle it, and your "set `custom` once and let it persist" workflow is the right answer.

The remaining argument for the alias — legacy TLP configs with `PLATFORM_PROFILE_ON_AC=balanced-performance` silently failing — is real but minor, and easily addressed by users updating their `tlp.conf` (or just leaving `platform_profile` to the BIOS, as I'm now doing). Not enough to justify the divergence from your intentional `custom` naming choice.

Closing this PR. Thanks for the thoughtful explanation and the references — I learned a lot, and I appreciate you taking the time to walk through how the various sysfs interfaces interact. I'll keep this commit on my personal fork for users who happen to want the alias behavior on this kind of hardware, but I won't push to bring it upstream.

PR #473, opened 2026-07-01
Add legion7 16iax7 k1cn

davyvalekestrel:


davyvalekestrel:
:)

st0nie:
The fancurve read-failure fallback in read_fancurve() returns a zeroed curve with err=0 when no cached curve exists. That makes a failed hardware read indistinguishable from a real all-zero curve — the GUI would display zeros and a subsequent apply could program a 0% curve to the EC (fans off -> overheat risk). Please propagate the read error when there is no valid cached curve (only fall back to the cached copy), so K1CN still works after a first successful write without masking genuine failures.

st0nie:
Fixed in 9f47513: read_fancurve() now propagates the hardware read error when no valid cached curve exists, instead of returning a zeroed curve with err=0. The cached-curve fallback (after a successful read/write) is unchanged, so K1CN keeps working. Merged, thanks!

Issue #486, opened 2026-08-08
[BUG] GPU OC and PPAB power limit read/write the wrong registers


chinese-wzq:
# Summary

On the 16IRX9 (BIOS NMCN) support added in #482, two GPU attributes appear to be **attached to the wrong WMI3 feature codes**:
- `gpu_oc` is attached to `GPU_POWER_BOOST` (`0x02010000`), which measures a power budget, not OC state
- `gpu_ppab_powerlimit` is attached to `POWER_TARGET_ON_AC_OFFSET_FROM_BASELINE` (`0x02040000`), which is a CPU-side value

Direct write tests on a Legion with BIOS **LPCN62WW** show these two codes behave the opposite way.

## Test 1 — `0x02010000` vs `0x02020000`, under FurMark load

| feature code written | value | measured GPU power | behaviour |
|---|---|---|---|
| `0x02020000` (base power limit / cTGP) | 90 | up to ~95 W | GPU follows the set limit |
| `0x02010000` (extra power budget / PPAB) | 10 (base limit kept at 80) | up to ~90 W | borrows power: **+10 W** above the base |

`0x02010000` is also a plain numeric register on my machine — I write 10 and read back 10 — i.e. a power-budget value, not an on/off OC state.

## Test 2 — `0x02040000`

| scenario | value = 55 | value = 90 |
|---|---|---|
| GPU-only load (FurMark) | GPU ~80 W | GPU ~80 W (no change) |
| CPU-only load (stress-ng) | CPU ~56 W | CPU ~56 W (no change) |
| CPU + GPU load | CPU gets ~56 W | CPU gets ~95–104 W |

`0x02040000` had no effect on GPU power draw; it only changed the *CPU* power share during combined load — consistent with "total processor power target on AC" being a CPU-side value, not a GPU extra-power budget.

# Impact

- `gpu_oc` never shows a real OC state; `echo 1 > gpu_oc` would write a power-budget value instead of enabling OC
- `gpu_ppab_powerlimit` would read/write the AC power target instead of the PPAB budget

For the record, the same `gpu_oc → GPU_POWER_BOOST` mapping also appeared independently in #464.

**Suggested fix**

- `gpu_ppab_powerlimit` → `GPU_POWER_BOOST`
- `gpu_oc` →  temporarily drop the WMI3 branch and revisit it once a better path is found.

[...]

PR #488, opened 2026-08-09
fix gpu oc and ppab

qquique:
- leave gpu_oc using only WMI DA7547F1-824D-405F-BE79-D9903E29CED7, some modern laptop models will see value 0,  functionality was removed from WMAD WMI Call.
- gpu_ppab_powerlimit points to GPU_POWER_BOOST(CPU to GPU Dynamic Boost) 0x02010000

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -934,6 +934,7 @@ static const struct model_config model_lpcn = {
 	.access_method_temperature = ACCESS_METHOD_WMI3,
 	.access_method_fancurve = ACCESS_METHOD_WMI3,
 	.access_method_fanfullspeed = ACCESS_METHOD_WMI,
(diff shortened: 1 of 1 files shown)


st0nie:
Thanks for fixing these mappings. The GPU OC/PPAB change makes sense, and the LPCN hardware results in issue #486 are helpful.

There is one thing I’d like to resolve before merging: `model_lpcn` is also used by the M3CN DMI entry. Adding `access_method_powerlimits = ACCESS_METHOD_WMI3` switches several CPU/GPU power-limit attributes for both LPCN and M3CN, not only PPAB on the tested laptop.

Could you either confirm the WMI3 power-limit path on M3CN hardware, or split M3CN into a separate model configuration that keeps the current access method? After that this should be straightforward to reconsider.

qquique:
I moved to its own model_m3cn, and removed the powerlimits flag so it stays as false.

PR #434, opened 2026-05-07
Fix ACPI access on Kernel 7.x virtual platform device

v4cachy:
**Problem:**
On Kernel >= 7.0.0, the virtual "legion" platform device (`platform_device_register_simple`) has no ACPI companion, leaving `priv->adev = NULL`. All ACPI-dependent features break:
- RapidCharge
- Battery conservation
- Fan readings
- CPU/GPU temperature

The original PR #423 tried `acpi_get_handle(NULL, path, &handle)` in `eval_int()`, but the default ACPI paths are relative (e.g. `"VPC0.GBMD"`) and don't resolve from the namespace root, so the lookup fails.

**Fix:**
Add `legion_find_ec_handle()` to locate the PNP0C09 Embedded Controller device in `acpi_init()`. This way `priv->adev` is set once at init and all existing code paths (`eval_int`, `exec_simple_method`) work unchanged.

**Changes:**
1. New helper `legion_find_ec_handle()` — tries common EC ACPI namespace paths
2. `acpi_init()` — finds PNP0C09 handle when virtual device provides NULL adev
3. `eval_int()` — reverted to plain `adev->handle` usage
4. `exec_simple_method()` — same revert

Fix by v4cachy

PR #513, opened 2026-08-21
fan_control.sh: remove hardcoded personal path

armoox:
## Problem

`fan_control.sh` could not run on any machine except its original author's:

1. `legion_cli` was invoked through a hardcoded private venv path (`/home/xuananh/Downloads/LenovoLegionLinux/.venv/bin/`), which fails everywhere else — and was redundant anyway, since the script already wrote `fan_fullspeed`/`fan_maxspeed` directly via sysfs right before
2. The sysfs base path was hardcoded to `PNP0C09:00`, which no longer exists on kernels >= 7.0 (interface moved to `platform:legion/legion`) — so even `status` failed there
3. `enable` wrote `powermode 2` (balanced) and `disable` wrote `powermode 3` (performance) — backwards relative to their labels

## Fix

- Auto-detect the sysfs base path: try the legacy `PNP0C09:00` layout first, then the kernel >= 7.0 `legion` layout — works on both
- Remove the foreign venv invocation and the duplicate write; direct sysfs writes are sufficient
- Drop the unrelated power-mode writes so the script does one job (force full fan speed on/off)
- Clear error message when the module interface is missing; `n/a` for unreadable attributes

## Testing

- Ran on Legion 5 15ACH6H (82JU), CachyOS kernel 7.2.0 (`legion` path): `status` reports correct fan RPMs and settings
- `bash -n` syntax check clean

### DIFF
--- a/fan_control.sh
+++ b/fan_control.sh
@@ -3,47 +3,71 @@
 # Legion Fan Control Script
 # Usage: ./fan_control.sh [enable|disable|status]
 
+LEGION_SYS="/sys/module/legion_laptop/drivers/platform:legion"
+
+find_base() {
+	local base
+	for base in "$LEGION_SYS/PNP0C09:00" "$LEGION_SYS/legion"; do
+		[ -d "$base" ] && printf '%s\n' "$base" && return 0
+	done
+	return 1
+}
+
+set_attr() {
+	echo "$2" | sudo tee "$1/$3" > /dev/null
+}
+
+read_attr() {
+	cat "$1/$2" 2>/dev/null || echo "n/a"
+}
+
+show_status() {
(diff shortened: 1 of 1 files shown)


armoox:
@st0nie Thank you for merging my 4 PR's in the main branch, Which are 

1. https://github.com/johnfanv2/LenovoLegionLinux/pull/434
2. https://github.com/johnfanv2/LenovoLegionLinux/pull/511
3. https://github.com/johnfanv2/LenovoLegionLinux/pull/512
4. https://github.com/johnfanv2/LenovoLegionLinux/pull/513

Special Thanks to @MrDuartePT for merging the PR#434 in the main branch

PR #517, opened 2026-08-25
feat(83sc): CD01-driven power limit clamping + discrete value snapping (PR-2-2)

Punch-Pain:
## PR-2-2: Capdata-Driven Power Limit Clamping

Builds on PR-2-1 (#500, merged) which added model/DMI wiring + `ACCESS_METHOD_WMI3_CLAMPED`.

### Motivation

On the LOQ 83SC the EC accepts any integer for these WMI features; out-of-range writes silently fail or misbehave instead of being rejected. Lenovo's Windows tooling never sends raw values — it clamps against ranges the BIOS publishes via WMI capability data. This PR brings the same behavior to Linux: **ranges are read from the hardware at probe time, nothing is hardcoded**, so BIOS updates that change limits are picked up automatically.

### Architecture

```
probe (legion_add)
 ├─ CD01 data blocks   (80 instances) → priv->capdata[]          min/max/step per feature per thermal mode
 ├─ DISCRETE data      (23 instances) → priv->discrete_features[] valid value sets for step==0 features
 └─ read_powermode()                  → priv->current_powermode CD01 rows are keyed by thermal mode

sysfs write → clamped_value()
 ├─ kstrtoint
 ├─ find CD01 row  (dev, feat, mode=current, type=0, supported&VALID)
 ├─ discrete set?  → snap to nearest value in set   (min/max NOT applied)
 └─ else step > 0? → snap to step, clamp to [min,max]
        → snprintf → wmi_common_method_other_store()   (unchanged upstream path)
```

Two WMI data blocks are cached at probe:

| Block | GUID | Format | Cached into |
|---|---|---|---|
| `LENOVO_CAPABILITY_DATA_01` | `7A8F5407-CB67-4D6E-B547-39B3BE018154` | buffer of `{id, supported, default, step, min, max}` | `capdata[80]` |
| `LENOVO_DISCRETE_DATA` | `91433B17-B7B7-4640-BB40-34C67349FBEC` | ACPI **package** `[IDs u32, Value u32]` per instance; same-ID instances form a set | `discrete_features[]` |

Attribute IDs encode `[31:24]=dev, [23:16]=feat, [15:8]=thermal mode, [7:0]=type`; matching compares the high half `(id >> 16) == ((u32)feature >> 16)` plus mode byte, type==0 and the VALID support bit.

### Decision logic (`capdata_clamp()`)

```
step > 0              → continuous : snap to step (DIV_ROUND_CLOSEST), clamp to [min,max]
step == 0 + known set → discrete   : snap to nearest value in set
step == 0, no set     → pass-through (BIOS validates)
```

Discrete snapping deliberately skips the min/max clamp: CD01's bounds lie for some discrete features — PPAB reports min=10 W while its DISCRETE_DATA set is `{0, 10, 15}`; applying min/max after the snap would break the 0 W option. Negative inputs are handled with int-typed comparisons (`clamp(value, (int)min, (int)max)`) and clamp to min, not max.

### Per-feature data (measured on 83SC)

| Feature | CD01 ID (custom mode) | Bounds / set | sysfs attribute(s) |
|---|---|---|---|
| CPU Short Term (SPPT) | `0x0101FF00` | 40–85 W, step 1 | `cpu_shortterm_powerlimit` |
| CPU Long Term (SPL) | `0x0102FF00` | 25–60 W, step 1 | `cpu_longterm_powerlimit` |
| CPU Temp limit | `0x0104FF00` | 85–100 °C, step 1 | `cpu_temperature_limit` |
| CPU Cross load | `0x0106FF00` | 20–30 W, step 1 | `cpu_cross_loading_powerlimit` |
| CPU TAU | `0x0107FF00` | discrete: 20,24,28,32,40,48,56,64,80,96,112,128,160 s | `cpu_l1_tau` |
| GPU PPAB / Dynamic Boost | `0x0201FF00` | discrete: 0, 10, 15 W | `gpu_ppab_powerlimit`, `gpu_oc` |
| GPU cTGP | `0x0202FF00` | discrete: 35, 40, 45, 50 W | `gpu_ctgp_powerlimit` |
| GPU Temp limit | `0x0203FF00` | 75–87 °C, step 1 | `gpu_temperature_limit` |
| GPU AC power target offset | `0x0204FF00` | 10–45 W, step 1 (continuous slider) | `gpu_power_target_offset` |
| GPU→CPU Dynamic Boost | `0x020BFF00` | discrete: 0, 10, 15 W | read-only attribute, no store path |

PL1/PL2 stores additionally run the existing coupling logic between clamp and write, preserving the PL1 ≤ PL2 invariant.

### Bug fixes included

1. **Wrong feature ID in `gpu_ppab_powerlimit` show/store** — the WMI3_CLAMPED path sent `GPU_POWER_TARGET_ON_AC_OFFSET_FROM_BASELINE` (0x0204) instead of `GPU_POWER_BOOST` (0x0201); reads/writes hit the wrong feature.
2. **Short-write return** — handlers now return the original user byte count (`err < 0 ? err : count`) rather than the clamped-string length; returning `strlen()` triggered sysfs short-write retries that made `echo | tee` exit non-zero on success.
3. **Signedness** — `min/max` are u32 while input is int; untyped comparison promoted negative writes to huge unsigned values (−50 clamped to *max*). Casted `clamp()` fixes negatives to clamp to *min*.

### Testing

Hardware-validated on Lenovo LOQ Essential 15IRX11 (83SC): all attributes above clamp/snap correctly across boundary, mid-range, and out-of-range inputs (69-case suite); PL1/PL2 coupling ON and OFF; negative inputs verified; discrete sets match Windows LLT option lists exactly.

### Scope

All new logic is gated behind `ACCESS_METHOD_WMI3_CLAMPED`, which only the 83SC model config selects. Upstream code paths are untouched below each guard; no new sysfs attributes, no behavior change for any other model. Diff: +247/−23 in `kernel_module/legion-laptop.c`.

### Related

- PR-2-1: #500 (merged) — model/DMI wiring
- PR-2-3: (planned) — keyboard backlight, OverDrive, other gaps


Punch-Pain:
@st0nie PR is ready for review. Summary of what landed since your last look:

- **CD01 capability data probing at probe time** — min/max/step read from the BIOS per feature per mode, no hardcoded tables
- **Discrete value snapping** from LENOVO_DISCRETE_DATA for step=0 features (PPAB, cTGP, TAU), matching LLT's StepperValue logic; discrete snap skips min/max so PPAB 0W stays valid
- **PL1/PL2 coupling** behind a cpu_pl_coupling attr (default on, model-gated)
- **Bug fixes**: gpu_ppab feature ID, short-write returns (err < 0 ? err : count), signedness (negatives clamp to min), explicit base-10 parsing
- **Scope unchanged**: everything gated to ACCESS_METHOD_WMI3_CLAMPED (83SC only), upstream paths verbatim, diff +247/-23

Hardware-validated on 83SC: 189-case audit suite (boundaries, INT_MIN/MAX, garbage input, format abuse, coupling matrix, reload persistence) — 0 failures, no kernel events. Full description with architecture diagrams in the PR body.

PR #528, opened 2026-09-02
fix: resolve 6 confirmed bugs plus code-quality cleanup

armoox:
## Summary

Fixes 6 confirmed bugs across the kernel module, Python package, CI workflows and deploy scripts, plus a set of code-quality improvements.

## Bug fixes

- **kernel-module**: fix null deref in feature lookup (`(!priv) && (priv->loaded)` → `||`); return `-EIO` instead of `-1` from sysfs store callbacks (`-1` surfaced as errno 255).
- **python**: `except BaseException` → `except Exception` (don't swallow `KeyboardInterrupt`/`SystemExit`); avoid `KeyError` when running as root without `SUDO_USER` set; correct `EnumSettingFeature.get()` return annotation (`bool` → `str`).
- **python/gui**: replace blocking `time.sleep` in controller readback with a deferred `QtCore.QTimer` readback so the UI thread is not frozen while reading hardware state.
- **ci**: fix `test_kenel_dkms_install.sh` → `test_kernel_dkms_install.sh` in all three workflow files (the Build step was silently failing).
- **deploy**: fix `build_containers.sh` working-dir bug; drop `curl --insecure` (disables TLS verification) from the deb download script.

## Code quality

- Replace remaining `print()` diagnostics with the configured logger.
- Fix internal identifier typos (`PowerProfilesDeamonService`→`Daemon`, `NotifcationSender`→`NotificationSender`, `get_ressource_path`→`get_resource_path`, `occured`→`occurred`, `seting`→`setting`) — no ABI impact.
- **legiond**: make GPU tool paths configurable via `nvidia_smi_path`/`rocm_smi_path` in `legiond.ini` (added `MAX_CMD_LEN`); replace `strcpy` with bounds-checked `snprintf` for the socket path.
- **smartfan**: quote hwmon paths and enable `set -u` in `setmyfancurve.sh`.

## Verification

- `checkpatch.pl`: 0 errors, 0 warnings on `legion-laptop.c`.
- `legiond` builds with `-std=gnu2x -Wall -Wextra`: zero warnings.
- `tests/test_python_cli.sh` passes; all Python compiles; bash scripts pass `bash -n`.

PR #530, opened 2026-09-04
fix(shell): fix shellcheck warnings across deploy/, tests/, smartfan

armoox:
Fixes shellcheck warnings found across all shell scripts in the repo. Suggested by @st0nie in #528.

**3 real bugs:**
- `deploy/python_uninstall_pkg.sh`: broken shebang `!/bin/bash` → `#!/bin/bash`
- `deploy/package_testing/download_install_debian.sh`: missing shebang entirely
- `tests/test_kernel_dkms_install.sh`: `cd` without `|| exit` — script would silently continue in the wrong directory on failure

**Code quality fixes:**
- Quote unquoted variables to prevent word splitting (SC2086)
- Replace deprecated `[ -a ]` test operator with `&&` (SC2166)
- Quote `7.2.2-1-cachyos` in apt-get install (SC2046)
- Split local declare+assign to avoid masking return values (SC2155)
- Use `read -ra` instead of array word splitting (SC2206)
- Add `-r` flag to `read` to preserve backslashes (SC2162)
- Remove unused `REPODIR` variable (SC2034)
- Suppress false-positive SC2016 for Makefile syntax in printf

```
shellcheck deploy/*.sh deploy/dependencies/*.sh deploy/dependencies/linux_kernel/*.sh deploy/package_testing/*.sh tests/*.sh extra/smartfan/*.sh
# EXIT: 0 — clean
```

### DIFF
--- a/deploy/build_kernelpatch.sh
+++ b/deploy/build_kernelpatch.sh
@@ -20,7 +20,7 @@ mkdir -p "${BUILD_DIR}"
 # Clone
 cd "${BUILD_DIR}"
 git clone --depth 1 --branch "v${KERNEL_VERSION}" https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
-cd ${BUILD_DIR}/linux
+cd "${BUILD_DIR}/linux"
 
 DRIVER_DIR="${BUILD_DIR}/linux/drivers/platform/x86"
 if [ -d "${DRIVER_DIR}/lenovo" ]; then
@@ -44,9 +44,10 @@ config LEGION_LAPTOP
 	  This is a driver for Lenovo Legion laptops and contains drivers for
 	  hotkey, fan control, and power mode.
 EOF
+# shellcheck disable=SC2016
 printf '\nobj-$(CONFIG_LEGION_LAPTOP) += legion-laptop.o\n' >> "${DRIVER_DIR}/Makefile"
 
-cd ${BUILD_DIR}/linux
+cd "${BUILD_DIR}/linux"
 git config user.name "John Martens"
 git config user.email "john.martens4@proton.me"
 git add --all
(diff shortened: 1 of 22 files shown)


armoox:
@st0nie following up on your suggestion from #528 — ran shellcheck over all shell scripts and did a manual static analysis pass on the kernel module (coccinelle/smatch aren't available without sudo).

This PR includes the fixes I'm confident about. A few findings I'm deferring because they need your input:

**1. `ecram_read` returns uninitialized value on failure (line 3119)**
On error, `value` is never assigned but still returned. Every caller treats it as valid — fan speeds, temps, power mode, fan curves all silently get garbage. Fixing this properly means changing the function signature and updating every caller across the file. Didn't want to make that call without checking with you first.

**2. Overwritten sensor values (lines 3689-3698)**
`ec_read_sensor_values` reads temps from model-specific register offsets, then immediately overwrites them with hardcoded `0xC5E6-C5E8`. The first three reads are dead code. Not sure which set is correct — model-specific or hardcoded — so I left it.

**3. Fancurve write functions reading past `fancurve->size` (lines 4316, 4435, 4535)**
`ec_write_fancurve_ideapad`, `_loq`, and `_legion2024` loop up to their hardware limit (8 or 10) without checking `fancurve->size`. If size is smaller, they read uninitialized points. Need to trace the callers to know if size is always guaranteed.

**4. Dead error checks in `ec_read_temperature`/`ec_read_fanspeed` (lines 3703, 3723)**
These have `if (!err)` that's always true because `err` is never updated. Connected to issue #1 — once `ecram_read` can report errors, these need to propagate them.

All of these involve hardware behavior or cascading changes that could break fan control on real laptops, so I'd rather not guess. Let me know how you want to handle them.

Issue #504, opened 2026-08-18
Support Request: Lenovo Legion 5 15AHP11 (83Q7)


Clyde6790c:
**Is your feature request related to a problem? Please describe.**
The `legion_laptop` kernel module requires `force=1` to load on the Lenovo Legion 5 15AHP11 (DMI Product Name: `83Q7`). When forced, direct Embedded Controller (EC) access fails, causing phantom LED initialization and backlight driver failures, though WMI monitoring and ACPI platform profile switching function properly.

**Describe the solution you'd like**
Official allowlist support and EC mapping for the Lenovo Legion 5 15AHP11 (`83Q7`), allowing proper WMI/EC initialization without requiring forced loading.

**Describe alternatives you've considered**
Reverting to Windows 11, which is undesirable.

**Model and Debug Info**
* **Laptop Model:** Lenovo Legion 5 15AHP11
* **DMI Product Name:** 83Q7
* **Installation Type:** Built via standard `make` (kernel module build)
* **Distribution:** Fedora 44 (KDE Plasma)
* **Kernel:** 7.1.8-200.fc44.x86_64
* **CPU:** AMD Ryzen 7 250 (16) @ 5.13 GHz
* **GPU Model 1:** NVIDIA RTX 5050 Laptop GPU
* **GPU Model 2:** AMD Radeon 780M Graphics
* **Keyboard Backlight:** RGB, controlled via Fn + Space
* **Lid/Logo Light:** None
* **IO-Port Light:** None

---

### Test Findings (`force=1` loaded with fallback profile `GKCN`)

* **ACPI Platform Profile Switching (Working):**
  * `cat /sys/firmware/acpi/platform_profile_choices` returns `low-power balanced performance`.
  * Writing to `/sys/firmware/acpi/platform_profile` successfully changes power limits and syncs the power button LED colors:
    * `low-power` (Blue LED)
    * `balanced` (White LED)
    * `performance` (Red LED)
* **WMI Monitoring (Working):**
  * Reads real-time CPU (43°C) and GPU (35°C) temperatures via WMI.
  * Reads live dual fan RPMs (`fan1`: 1300 RPM, `fan2`: 1600 RPM) via WMI3 / `hwmon`.
  * Successfully reads 10 hardware fan curve points via WMI.
* **Direct EC / Backlight (Failing):**
  * Direct EC register access returns `0` or errors (`-5`), indicating an unmapped EC layout.
[...]

PR #531, opened 2026-09-04
feat(kernel-module): add T3CN (Legion 5 15AGP11, 83Q6) using model_t2cn

arcvrx:
Adds a DMI allowlist entry for the **Legion 5 15AGP11 (83Q6, AMD Ryzen AI 9 465 + RTX 5070)**, BIOS prefix `T3CN`, pointing at the existing `model_t2cn`.

This is the same platform generation as the 83Q7 added in the T2CN entry — EC chip `0x5509`, vendor EC RAM reads blocked by firmware, fans/temps/fancurve over WMI3 — so no new `model_config` is needed.

### The problem

The 83Q6 has no DMI entry, so the module either refuses to load, or with `force=1` falls back to `optimistic_allowlist[0]` (GKCN / `model_v0`, a 2021 Legion 7 config) and reads the wrong EC offsets. That fails silently: sensors return 0 rather than erroring.

```
Using configuration for system: GKCN
Skipped checking embedded controller id
Fan 1: 0 RPM   Fan 2: 0 RPM
CPU Temperature: +0.0°C   GPU Temperature: +0.0°C
Error WMI call for reading brightness: expected a value between 1 and 3, but got 0
```

### After

```
is_denied: 0; is_allowed: 1; do_load_by_list: 1; do_load: 1
Using configuration for system: T3CN
Read embedded controller ID 0x5509
Keyboard backlight handling disabled by this driver
```

Matched by the allowlist rather than forced, and the EC ID check now **passes on its own** instead of being skipped.

### Fan verification

Sampled against the independent `yogafan` hwmon under 20-thread CPU load:

| load | `legion_hwmon` | `yogafan` | CPU temp |
|---|---|---|---|
| idle | 0 RPM | 0 RPM | 38°C |
| 10s | 1200 RPM | 1250 RPM | 75°C |
| 20s | 2000 RPM | 1900 RPM | 79°C |
| 40s | 2500 RPM | 2800 RPM | 84°C |
| 60–90s | 3100 RPM | 3300 RPM | 85–86°C |
| cooling | 2600 RPM | 2700 RPM | 47°C |

CPU temperature tracked 38°C idle → 86°C loaded → 47°C on release. Both fans report (fan1 1950 / fan2 2050 RPM while cooling). Offsets between the two sources are sampling skew, not disagreement.

### Why `model_t2cn` verbatim

Every model-specific flag was checked against the hardware:

| Field | Evidence on 83Q6 |
|---|---|
| `embedded_controller_id = 0x5509` | `Read embedded controller ID 0x5509`, check passes unforced |
| `access_method_keyboard = NO_ACCESS` | brightness read fails identically to 83Q7 (the reason for ea15282) |
| `skip_ylogo_light = true` | no lid lighting on this chassis |
| `skip_ioport_light = true` | no illuminated rear I/O ports |
| `access_method_fanspeed/temperature = WMI3` | table above |

### Known limitations

Unchanged by this patch and shared with the 83Q7: IC temperature (`temp3`) still reads 0, and the EC-backed power-limit attributes still read 0, since vendor EC RAM access is blocked by firmware.

Also worth noting: the mainline `lenovo-wmi-other` driver logs `fan reporting/tuning is unsupported on this device` for `DC2A8805-…` on this machine, yet LLL's WMI3 path reads fans correctly — so that probe doesn't predict WMI3 usability here.

### Testing

- Machine: Lenovo Legion 5 15AGP11 (83Q6), BIOS T3CN37WW, board LNVNB161216
- Kernel 7.2.2, Artix Linux
- `make` clean; `checkpatch.pl --ignore LINUX_VERSION_CODE --ignore CONSTANT_COMPARISON`: **0 errors, 0 warnings**
- Loaded via `make reloadmodule` (plain `insmod`, no `force`) to confirm the DMI match binds on its own

Refs #504

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -2061,6 +2061,17 @@ static const struct dmi_system_id optimistic_allowlist[] = {
 		.driver_data = (void *)&model_t2cn
 	},
 	{
+		// e.g. Legion 5 15AGP11 (83Q6, AMD Ryzen AI 9 465 + RTX 5070) -
+		// same platform as the 83Q7 above; EC 0x5509, fancurve via
+		// WMI3 (#504)
+		.ident = "T3CN",
+		.matches = {
+			DMI_MATCH(DMI_SYS_VENDOR, "LENOVO"),
+			DMI_MATCH(DMI_BIOS_VERSION, "T3CN"),
+		},
+		.driver_data = (void *)&model_t2cn
+	},
+	{
 		// e.g. Legion 5 15IRX10 (83LY)
 		.ident = "QNCN",
 		.matches = {


toad6115149:
**Confirmed persistent across reboot, with `force=1` removed.**

Installed via DKMS and rebooted, with `options legion_laptop force=1` commented out of `/etc/modprobe.d/legion-laptop.conf` so the allowlist is the only thing binding it:

```
[    9.829351] legion legion: is_denied: 0; is_allowed: 1; do_load_by_list: 1; do_load: 1
[    9.829352] legion legion: Using configuration for system: T3CN
[   12.099475] legion legion: Read embedded controller ID 0x5509
```

Note what's absent: no `legion_laptop is forced to load` and no `Skipped checking embedded controller id`. The EC ID check passes on its own rather than being bypassed.

Sensors at idle after boot:

```
legion_hwmon-isa-0000
Fan 1:           1700 RPM
Fan 2:           2000 RPM
CPU Temperature:  +39.0°C
GPU Temperature:  +35.0°C
IC Temperature:    +0.0°C
```

(`IC Temperature` reading 0 is the known `temp3` limitation noted in the description, unchanged by this patch.)

PR #532, opened 2026-09-05
fix(kernel-module): fix 18 static-analysis findings in legion-laptop.c

armoox:
@st0nie following up on your suggestion from #528 — ran a manual static-analysis pass on the kernel module (coccinelle/smatch weren't available). This PR contains the fixes I'm confident about. The **deferred** findings below are ones that involve hardware behavior or cascading changes that could break fan control, so I'd rather not guess — they need your input.

---

### Fixed in this PR

**High severity:**
- `acpi_read_fanspeed` (line 3765): fan 0 used fan 2's ACPI path (copy-paste); fan 1 wasn't handled at all — falls into `-EINVAL`
- `legion_wmi_notify` (line 7007): dereferenced `priv->ppdev` after `mutex_unlock` — use-after-unlock on kernel >= 6.14
- `legion_add` (line 8665): `acpi_init()` return value was dropped on kernel >= 7.0 — kept probing with broken ACPI access

**Medium severity:**
- `acpi_read_powermode` (line 4947): wrote uninitialized value to `*powermode` before checking error
- `ecram_memoryio_read/write` (lines 2925, 2943): added missing upper-bound check — could read/write past the mapped EC region
- 4x `sprintf()` → `sysfs_emit()` in sysfs show methods (lines 7366, 7422, 7534, 7704) per kernel sysfs API requirements

**Low severity:**
- 4x wrong `-ENOMEM` → `-EINVAL`/`-ENODEV`/`-EIO` (lines 4934, 4973, 8530, 8696)
- 3x removed unreachable `return err;` after exhaustive switch (lines 6584, 6615, 6647)

Build verified: `make CC=clang LD=ld.lld LLVM=1` clean.

---

### Deferred — need your input

**1. `ecram_read` returns uninitialized value on failure (line 3119)**
On error, `value` is never assigned but still returned. Every caller treats it as valid — fan speeds, temps, power mode, fan curves all silently get garbage. Fixing this properly means changing the function signature to return an error and updating every caller across the file, which is a large change.

**2. Overwritten sensor values (lines 3689-3698)**
`ec_read_sensor_values` reads temps from model-specific register offsets, then immediately overwrites them with hardcoded `0xC5E6-C5E8`. The first three reads are dead code. Not sure which set is correct — model-specific or hardcoded — so I left it.

**3. Fancurve writes reading past `fancurve->size` (lines 4316, 4435, 4535)**
`ec_write_fancurve_ideapad`, `_loq`, `_legion2024` loop up to their hardware limit (8 or 10) without checking `fancurve->size`. If size is smaller, they read uninitialized points and write them to the EC. Need to trace callers to know if size is always guaranteed.

**4. Dead error checks in `ec_read_temperature`/`ec_read_fanspeed` (lines 3703, 3723)**
These have `if (!err)` that's always true because `err` is never updated. Connected to issue #1 — once `ecram_read` can report errors, these need to propagate them.

All four involve real hardware or large cascading changes. Let me know how you want to handle them.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -2925,8 +2925,9 @@ static void ecram_memoryio_exit(struct ecram_memoryio *ec_memoryio)
 static ssize_t ecram_memoryio_read(const struct ecram_memoryio *ec_memoryio,
 				   u16 ec_offset, u8 *value)
 {
-	if (ec_offset < ec_memoryio->physical_ec_start) {
-		pr_info("Unexpected read at offset %d into EC RAM\n",
+	if (ec_offset < ec_memoryio->physical_ec_start ||
+	    ec_offset - ec_memoryio->physical_ec_start >= ec_memoryio->size) {
+		pr_info("Unexpected read at offset 0x%x into EC RAM\n",
 			ec_offset);
 		return -EIO;
 	}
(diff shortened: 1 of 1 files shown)


st0nie:
LGTM

armoox:
> LGTM

@st0nie What, what you mean by LGTM? are you referring to (Looks Good To Me)
i asked you to check the deferred  Section, cause i need your input on that.

st0nie:
Oops, I only review the diffs you provided in this PR. I will answer you as soon as I am free.

armoox:
@st0nie if you think i should ignore the deferred section on the PR note, let me know i'll ignore it or if you want to fix it i can do it or you can do it if you want, so yes get back to to me for this issue, I would appreciate it and also tell me what to do about it so i can know what should i do with the deferred section

Issue #491, opened 2026-08-14
EC Chip ID 0x5508 — fancurve read/write consistently broken across multiple models


Silver-Rust-18:
Opening this to consolidate a pattern showing up across several separate issues/PRs, all sharing EC Chip ID 0x5508 (EC Version 2b0) — custom fan curve access appears broken across this entire EC chip generation, regardless of which access method is used.

Affected reports so far:
- #337 (Legion Pro 5i 16IAX10H) — "test fan curves values does not match those in the README"
- #359 / #475 (Legion 5 15IAX10, 83F0) — confirmed EC3 fancurve read returns garbage (reversed temp ranges, impossible speeds); WMI3 FAN_GET_TABLE returns a real EPERM from the ACPI call itself
- #374 (LOQ 15IRX10, 83JE, R3CN) — source of the EC3 config later found broken elsewhere
- #384 (LOQ 15AHP10, 83JG, R8CN) — title literally "EC locked; fancurve EC=5508"
- #400 (Legion Pro 5 16ADR10, 83LT, AMD) — same EC chip despite different CPU vendor
- #414 (Legion 5 15IRX10, 83LY, QNCN) — fancurve via WMI returns 10 zeroed points
- #476 (Legion 7 16IAX10, 83KY, RXCN, unmerged) — proposed WMI3-based fancurve method; untested against most of the above models. Worth flagging that this PR's own testing claims were internally inconsistent (references both 16IAX10 and 16IRX9 as the tested device), so I'd treat its config as an unverified lead rather than confirmed data.

Common thread across all of these: it's specifically the custom fan curve table — read and/or write — that fails, across both EC3 and WMI3 access methods, on every model reporting this chip ID. Which access method works for basic monitoring (temperature/fan RPM) varies by board — WMI-based reads worked for #359/#475 and #384, but #400's own data shows EC-based GPU temp working (87°C, plausible) while WMI3 returned 0 for the same reading — so this isn't a uniform "WMI good, EC bad" split, just a fan curve that's consistently broken specifically.

Given how consistent the fancurve failure is across five+ different chassis (Intel and AMD, Legion 5/7/Pro, LOQ), this looks less like a per-model config gap and more like either:
1. A genuine firmware-level lock on custom fan curve writes for this EC generation, or
2. A different access mechanism entirely (not EC3 or WMI3) that none of these configs have found yet

[...]

Issue #535, opened 2026-09-07
Fan curve writes via sysfs have no effect on LOQ 15IRX10 (83JE, EC 5508)


aztrixx:
## Description

I have a Lenovo LOQ 15IRX10 (Type **83JE**) running CachyOS, kernel `7.2.2-1-cachyos`. `legion_laptop` loads successfully with `force=1` and correctly reads CPU/GPU temps, fan RPM, and powermode — but **writes to the fan curve (`pwm*_auto_pointN_temp`, `pwm*_auto_pointN_pwm`) via the hwmon sysfs interface have no effect on the hardware**, even after switching to the `custom` powermode (3) and repeated writes.

## Hardware

- Model: Lenovo LOQ 15IRX10, Type 83JE
- BIOS: R3CN44WW
- EC Chip ID: **5508**, EC Chip Version: 2b0
- CPU: Intel i5-13450HX, GPU: RTX 5050 (Blackwell)
- Distro: CachyOS (Arch-based), kernel 7.2.2-1-cachyos (clang-built)

## Steps to reproduce

1. Compile with `make CC=clang LD=ld.lld` (required — CachyOS kernel is clang-built, gcc fails on clang-specific flags)
2. `sudo make forcereloadmodule` (or installed via DKMS with the same CC/LD flags) — loads cleanly:
   ```
   legion_laptop: Read identifying information: DMI_SYS_VENDOR: LENOVO; DMI_PRODUCT_NAME: 83JE; DMI_BIOS_VERSION:R3CN44WW
   legion legion: legion_laptop is forced to load.
   legion legion: legion_laptop loaded for this device
   ```
3. Set powermode to custom: `echo 3 | sudo tee /sys/devices/platform/legion/powermode` → confirmed `cat` returns `3`
4. Write new fan curve values directly via sysfs, e.g.:
   ```
   echo 100 | sudo tee /sys/class/hwmon/hwmon5/pwm1_auto_point10_temp
   echo 255 | sudo tee /sys/class/hwmon/hwmon5/pwm1_auto_point10_pwm
   ```
   (writes succeed with no error, as long as `_accel`/`_decel` fields are left untouched — see below)
5. Check `sudo cat /sys/kernel/debug/legion/fancurve` — the table does **not** reflect the new values; it stays close to the stock curve regardless of how many times the script is re-run or whether the system is rebooted between attempts.

## Additional finding: writing acceleration/deceleration crashes with EIO

[...]

PR #539, opened 2026-09-07
fix(kernel): five fixes for LOQ Essential 15IRX11 (83SC) - fancurve, fan_fullspeed, phantom fan2, fan_max, platform_profile

### LINKED ISSUE #538: [BUG] LOQ Essential 15IRX11 (83SC): fancurve reads zeros, fan_fullspeed no-op, phantom fan2, wrong fan1_max, platform_profile fails in powermode 224
model: LOQ Essential 15IRX11, DMI 83SC, BIOS SECN14WW, EC 0x5508, kernel 6.18.48
specs: i7-13650HX (6P+8E, 14C/20T) RTX 5050 8GB, 16GB DDR5
the model ships with couple of configurations (i5-13450HX or i7-13650HX, RTX 5050 or 5060, both GPUs at 65W tho) and the power numbers below are from the i7 one, the single fan is probably the same across all of them since it's the same chassis

force=1 is not needed btw, the DMI match works, model_secn picks it up fine

found 5 things while chasing a stutter in CS2, all reproduced on hardware, patches for all of them at the bottom

## 1. fan curve reads all zeros

every pwmN_auto_pointM_temp reads 0 so the curve looks broken. the WMI3 path reads FAT2 which has no temperature data in it at all, FTS0-9 and FSS0-9 are just duplicate copies of the same speed array

model_secn already declares ec_register_offsets_loq_v1 which is the same table model_r3cn uses (LOQ 15IRX10, same EC 0x5508) and that one reads over EC3 and gets temps fine, so this looks like it was just pointed at the wrong access method when the model got added

before: `pt1 pwm=2 cpu=0` `pt2 pwm=5 cpu=0`
after: `pt1 pwm=0 cpu=75` `pt2 pwm=86 cpu=78` `pt3 pwm=102 cpu=80` `pt4 pwm=109 cpu=85`

decodes to 0/3400/4000/4300 rpm at 75/78/80/85C which matches the FNT0 table in the laptop's own ACPI dump exactly. raw EC bytes confirm the (min_temp, max_temp, rpm/100) triples:

```
+107: 00 4b 00   ->  --  75C     0 rpm
+10a: 3c 4e 22   ->  60  78C  3400 rpm
+10d: 40 50 28   ->  64  80C  4000 rpm
+110: 43 55 2b   ->  67  85C  4300 rpm
```

only tested on this one machine, if WMI3 was picked on purpose for some other 83SC config then that's worth knowing

## 2. fan_fullspeed silently does nothing outside powermode 255

in powermode 3, writing 1 succeeds and reads back 1, fan stays at 2600rpm. in powermode 255 the same write pins it to 4300rpm, checked against 3700rpm with it off at the same temperature

the firmware takes the WMI call and stores the flag, the fan controller just ignores it outside custom mode, so readback can't catch this because the flag really is set, only a powermode check works. reporting success on a write that does nothing sent me down a long dead end

## 3. phantom fan2

this chassis has one physical fan (opened it multiple times, changed paste twice) but hwmon shows fan2_input, fan2_label, fan2_target

60 samples during a thermal ramp under auto control, RPM moving 3000 -> 2000 -> 3400 through about 15 distinct values, fan1_input equalled fan2_input at all 60 samples, 0 divergence. two real fans would drift at least a bit

firmware ships 15 fan tables split across fanid=1 (CPU sensor 4) and fanid=2 (GPU sensor 5), so it's a real two fan design shared across the chassis family and the LOQ Essential is the single fan variant

note: fan2_max has to stay visible, legion_gui reads it with no existence check so hiding it makes the whole fan curve table show zeros (found that the hard way). pwm2_auto_point* left alone too since that's the gpu sensor curve not fan2 telemetry

## 4. fan1_max says 10000 rpm, real ceiling is 5400

every firmware table caps at 4300 and FTTD declares 5400 as the hardware max (seen it hit on windows, rarely)

MAX_RPM is also used for the RPM_HUNDRED pwm conversion, so changing only the reported max breaks the round trip, kernel encodes with one number and userspace decodes with another. legion_gui ended up showing 1900/2200/2400 rpm where the real values were 3400/4000/4300. so max_rpm goes in struct fancurve and both sides use it

## 5. platform_profile read fails in powermode 224 on kernels < 6.19

PLATFORM_PROFILE_MAX_POWER doesn't exist before 6.19 so the case is compiled out, powermode 224 falls into `default: return -EINVAL` and reading platform_profile just fails with nothing

```
echo 224 | sudo tee /sys/bus/platform/devices/PNP0C09:00/powermode
cat /sys/firmware/acpi/platform_profile     # nothing
```

## not bugs, checked and dropped

- platform_profile rejecting `custom` is correct, kernel docs say custom is driver reported status not something you write. writing 255 to powermode is the right way and platform_profile then reads custom
- fan curve temperatures can't be written by anyone including windows. SFAN declares ten temperature words (F00F-F018), a sensor id and a sensor length, and never reads any of them, each appears exactly once across all 238 lines of the method. thresholds come from internal packages FI00-FI09 picked by power mode. also F003-F00C are indices into a preset speed table not rpm (`CRP0 = Local0[F003 + 2] / 100`)
- ppt_pl1_spl / ppt_pl2_sppt returning EINVAL is lenovo-wmi-other not this driver, and doesn't matter since RAPL does the same job and works

## also, not a driver bug but worth knowing for this model

stock PL1 is 55W sustained on a single fan cooler whose fan maxes at 4300rpm by 85C. CPU reaches TCC at 97C (Tjmax 100, offset 3) and duty cycles via T-states, which userspace samples as ~400MHz, below the 800MHz scaling_min_freq, so it looks like a cpufreq/governor bug and isn't. no governor setting can prevent it

dropping PL1 to 45W and the window from 56s to 8s gives 70-72C with zero PROCHOT under a 14-thread all core load, and 0 events across 240s of combined CPU+GPU. done entirely through mainline intel_rapl, MSR_PKG_POWER_LIMIT bit 63 is clear so the BIOS doesn't lock power limits on this model

a sane per model default for single fan variants would fix this for everyone with this laptop

## patches

all 5 build clean with LLVM=1 and pass tests/test_kernel_checkpath.sh, verified on hardware

<details>
<summary>diff against main</summary>

```diff
diff --git a/kernel_module/legion-laptop.c b/kernel_module/legion-laptop.c
index 9e4c105..922b646 100644
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -257,6 +257,9 @@ struct model_config {
 	/* fan_target registers hold duty-cycle (0-100); scale by 100 to approximate RPM */
 	bool fan_target_is_duty;
 	bool has_four_fans;
+	bool has_single_fan;
+	u16 fan_max_rpm;
+	bool fanfullspeed_requires_custom_powermode;
 	bool skip_ylogo_light;
 	bool skip_ioport_light;
 	// EC register holding the Y-Logo light state; 0 = not available
@@ -1594,7 +1597,7 @@ static const struct model_config model_secn = {
 	.access_method_keyboard = ACCESS_METHOD_WMI2,
 	.access_method_fanspeed = ACCESS_METHOD_WMI3,
 	.access_method_temperature = ACCESS_METHOD_WMI3,
-	.access_method_fancurve = ACCESS_METHOD_WMI3,
+	.access_method_fancurve = ACCESS_METHOD_EC3,
 	.access_method_fanfullspeed = ACCESS_METHOD_WMI3,
 	.access_method_powerlimits = ACCESS_METHOD_WMI3_CLAMPED,
 	.acpi_check_dev = false,
@@ -1615,6 +1618,9 @@ static const struct model_config model_secn = {
 	.has_fan_unlock = true,
 	.has_fn_lock = true,
 	.has_flip_to_start = true,
+	.has_single_fan = true,
+	.fan_max_rpm = 5400,
+	.fanfullspeed_requires_custom_powermode = true,
 };
 
 // Legion 5i Gen 10 (83VK) - 2025/2026, Intel Arrow Lake-HX + RTX 5060
@@ -3250,6 +3256,7 @@ static const struct fancurve_point fancurve_point_zero = { 0, 0, 0, 0, 0,
 struct fancurve {
 	struct fancurve_point points[MAXFANCURVESIZE];
 	enum fan_speed_unit fan_speed_unit;
+	u16 max_rpm;
 	// number of points used; must be <= MAXFANCURVESIZE
 	size_t size;
 	// the point at which fans are run currently
@@ -3300,11 +3307,14 @@ static bool fancurve_set_speed_pwm(struct fancurve *fancurve, int point_id,
 	case FAN_SPEED_UNIT_PWM:
 		*speed = clamp_t(u8, value, 0, 255);
 		return true;
-	case FAN_SPEED_UNIT_RPM_HUNDRED:
+	case FAN_SPEED_UNIT_RPM_HUNDRED: {
+		u32 max_rpm = fancurve->max_rpm ? fancurve->max_rpm : MAX_RPM;
+
 		*speed = clamp_t(
-			u8, (value * MAX_RPM + (100 * 255) - 1) / (100 * 255),
+			u8, (value * max_rpm + (100 * 255) - 1) / (100 * 255),
 			0, 255);
 		return true;
+	}
 	default:
 		pr_info("No method to set for fan_speed_unit %d.",
 			fancurve->fan_speed_unit);
@@ -3335,9 +3345,12 @@ static bool fancurve_get_speed_pwm(const struct fancurve *fancurve,
 	case FAN_SPEED_UNIT_PWM:
 		*value = speed;
 		return true;
-	case FAN_SPEED_UNIT_RPM_HUNDRED:
-		*value = speed * 255 * 100 / MAX_RPM;
+	case FAN_SPEED_UNIT_RPM_HUNDRED: {
+		u32 max_rpm = fancurve->max_rpm ? fancurve->max_rpm : MAX_RPM;
+
+		*value = speed * 255 * 100 / max_rpm;
 		return true;
+	}
 	default:
 		pr_info("No method to get for fan_speed_unit %d.",
 			fancurve->fan_speed_unit);
@@ -4390,6 +4403,7 @@ static int ec_read_fancurve_loq(struct ecram *ecram,
 	size_t struct_offset_ecramsys = 6;
 
 	fancurve->fan_speed_unit = FAN_SPEED_UNIT_RPM_HUNDRED;
+	fancurve->max_rpm = model->fan_max_rpm;
 	for (i = 0; i < FANCURVESIZE_LOQ; ++i) {
 		struct fancurve_point *point = &fancurve->points[i];
 
@@ -6693,6 +6707,27 @@ static ssize_t fan_fullspeed_show(struct device *dev,
 	return sysfs_emit(buf, "%d\n", state);
 }
 
+static int fanfullspeed_write_allowed(struct legion_private *priv, bool state)
+{
+	int powermode;
+	int err;
+
+	if (!state || !priv->conf->fanfullspeed_requires_custom_powermode)
+		return 0;
+
+	err = read_powermode(priv, &powermode);
+	if (err)
+		return err;
+
+	if (powermode != LEGION_WMI_POWERMODE_CUSTOM) {
+		pr_info("fan_fullspeed needs powermode %d (custom), current is %d\n",
+			LEGION_WMI_POWERMODE_CUSTOM, powermode);
+		return -EBUSY;
+	}
+
+	return 0;
+}
+
 static ssize_t fan_fullspeed_store(struct device *dev,
 				   struct device_attribute *attr,
 				   const char *buf, size_t count)
@@ -6707,6 +6742,8 @@ static ssize_t fan_fullspeed_store(struct device *dev,
 
 	mutex_lock(&priv->fancurve_mutex);
 	err = fan_control_write_allowed(priv);
+	if (!err)
+		err = fanfullspeed_write_allowed(priv, state);
 	if (!err)
 		err = write_fanfullspeed(priv, state);
 	mutex_unlock(&priv->fancurve_mutex);
@@ -7159,11 +7196,13 @@ static int legion_platform_profile_get(struct platform_profile_handler *pprof,
 		*profile = PLATFORM_PROFILE_CUSTOM;
 		break;
 #endif
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 19, 0)
 	case LEGION_WMI_POWERMODE_MAX_POWER:
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 19, 0)
 		*profile = PLATFORM_PROFILE_MAX_POWER;
-		break;
+#else
+		*profile = PLATFORM_PROFILE_PERFORMANCE;
 #endif
+		break;
 	default:
 		return -EINVAL;
 	}
@@ -7462,6 +7501,11 @@ static struct attribute *sensor_hwmon_attributes[] = {
 static ssize_t fan_max_show(struct device *dev,
 			    struct device_attribute *devattr, char *buf)
 {
+	struct legion_private *priv = dev_get_drvdata(dev);
+
+	if (priv && priv->conf->fan_max_rpm)
+		return sprintf(buf, "%d\n", priv->conf->fan_max_rpm);
+
 	return sprintf(buf, "%d\n", MAX_RPM);
 }
 
@@ -8117,6 +8161,11 @@ static umode_t legion_hwmon_sensor_is_visible(struct kobject *kobj,
 	    attr == &sensor_dev_attr_fan4_label.dev_attr.attr)
 		supported = supported && priv->conf->has_four_fans;
 
+	if (attr == &sensor_dev_attr_fan2_input.dev_attr.attr ||
+	    attr == &sensor_dev_attr_fan2_label.dev_attr.attr ||
+	    attr == &sensor_dev_attr_fan2_target.dev_attr.attr)
+		supported = supported && !priv->conf->has_single_fan;
+
 	return supported ? attr->mode : 0;
 }
 
```

</details>

blussyya:
Fixes #538

five fixes for the LOQ Essential 15IRX11 (83SC), all found while chasing a CS2 stutter that turned out to be thermal. every one is reproduced on hardware, details and repro commands are in #538

- **fancurve reads all zeros** - `model_secn` used `ACCESS_METHOD_WMI3`, and that path reads `FAT2` which has no temperature data at all (`FTS0-9` and `FSS0-9` are duplicate copies of the same speed array). it already declares `ec_register_offsets_loq_v1`, the same table `model_r3cn` uses with EC3, so switching to `ACCESS_METHOD_EC3` makes it read real trip temps and rpm. this is the one line that matters most
- **fan_fullspeed** silently no-ops outside powermode 255, write succeeds and reads back 1 but nothing happens. now returns `-EBUSY`, gated per model
- **phantom fan2** hidden behind a new `has_single_fan`, following the `has_four_fans` pattern. `fan2_max` is deliberately left visible because legion_gui reads it with no existence check and hiding it blanks the whole fan curve table. `pwm2_auto_point*` is left alone too since that's the gpu sensor curve, not fan2 telemetry
- **fan1_max** was the global `MAX_RPM` (10000), real ceiling on this chassis is 5400. `MAX_RPM` also drives the RPM_HUNDRED pwm conversion so changing only the reported max breaks the round trip, `max_rpm` now lives in `struct fancurve` and both sides use it, falling back to `MAX_RPM` when a model doesn't set one
- **platform_profile** read returned `-EINVAL` in powermode 224 on kernels below 6.19 because `PLATFORM_PROFILE_MAX_POWER` doesn't exist there. maps to `PLATFORM_PROFILE_PERFORMANCE` on older kernels, 6.19+ untouched

## verified

built with `make LLVM=1` (kernel is clang+LTO), `tests/test_kernel_checkpath.sh` passes, 0 errors 0 warnings same as the baseline

on hardware after the patches:

```
fan1_max                    5400          (was 10000)
pwm1_auto_point1..4_temp    75 78 80 85   (was 0 0 0 0)
pwm1_auto_point1..4_pwm     0 86 102 109  -> 0/3400/4000/4300 rpm
fan2_input                  gone
platform_profile @ pm 224   performance   (was empty)
fan_fullspeed @ pm 3        EBUSY         (was fake success)
```

the temps and rpm match the `FNT0` table in the machine's own ACPI dump exactly

## caveat

single machine. the 83SC ships in a few configs (i5-13450HX or i7-13650HX, RTX 5050 or 5060) and i only have the i7/5050 one. if `WMI3` was chosen deliberately for another variant then the first commit needs narrowing, happy to change it


blussyya:
adding context after reading #491, since it argues against the reasoning i used for the first commit and i'd rather raise it myself

#491 consolidates fancurve breakage across every EC 0x5508 model and specifically notes that the r3cn EC3 config was "later found broken elsewhere", and there's a comment there warning that reusing another model's ec_register_offsets is exactly what produces garbage curves. that's the same reasoning i leaned on ("model_secn already declares ec_register_offsets_loq_v1, same as model_r3cn")

so to be clear about what is and isn't evidence here:

the values EC3 returns on the 83SC are not self-reported plausible numbers, they match the machine's own ACPI FNT0 table exactly. decompiled from the laptop's SSDT4, mode 224, cpu sensor:

```
temp C :  75    78    80    85   ...
fan rpm :  0  3400  4000  4300   ...
```

and the raw EC bytes decode to the same thing as (min_temp, max_temp, rpm/100) triples:

```
+107: 00 4b 00   ->  --  75C     0 rpm
+10a: 3c 4e 22   ->  60  78C  3400 rpm
+10d: 40 50 28   ->  64  80C  4000 rpm
+110: 43 55 2b   ->  67  85C  4300 rpm
```

that's an independent source agreeing with the driver read, which is different from the "reversed temp ranges, impossible speeds" failure mode #491 describes for 83F0. so on this board the loq_v1 offsets are correct, whatever is true elsewhere

writes work too, and not just at the sysfs level. after writing a curve that ramps from 45C the fan tracked it live (71C -> 2500rpm, 74C -> 2800, 80C -> 3000) where the stock curve is 0 rpm below 75C, so hardware behaviour actually changed. worth noting because #535 reports writes having no effect on 83JE which is the same EC and the neighbouring model

what i am **not** claiming: that this fixes #491, #535, or any other 0x5508 model. one board, verified against its own firmware. if the maintainers would rather see an ec_register_offsets_secn created instead of reusing loq_v1, that's easy to do even though the values come out identical here, just say which you prefer

the other four commits are independent of all this

armoox:
@blussyya Is this AI Generated? just curious, No offense tho

blussyya:
@armoox 
no offense taken dw
yeah the fixes were done by claude and i made it push this, not gonna claim ive done all the work

st0nie:
Thanks @blussyya for this thorough, hardware-verified fix set — really nice detective work tracing the CS2 stutter back to thermal behavior, and the write-up in #538 is excellent.

Merged. A few notes:

- The changes are all properly gated behind `model_secn`-specific flags (`has_single_fan`, `fan_max_rpm`, `fanfullspeed_requires_custom_powermode`), so other models are unaffected — exactly how we want hardware-affecting changes handled.
- I rebased the branch onto latest `main` to resolve a small conflict in `fan_max_show()`: #542 had meanwhile converted it to `sysfs_emit()`, so the resolution keeps your per-model `fan_max_rpm` logic but uses `sysfs_emit()` instead of `sprintf()`. The rebased branch was pushed back here (maintainer edit), CI is green, and the kernel module builds cleanly.
- The `access_method_fancurve` WMI3 → EC3 switch is the headline fix; good catch that `FAT2`'s FTS/FSS fields are duplicate speed arrays with no real temperature data.

Thanks again — fixes like this are exactly what keeps LLL useful on newer hardware!

PR #542, opened 2026-09-08
fix(kernel-module): fix 5 static-analysis findings in legion-laptop.c

armoox:
## Summary

Fixes 5 findings in `kernel_module/legion-laptop.c` found with the clang static analyzer and a review of the sysfs/debugfs show paths. All are non-behavioral cleanups or error-path fixes — identical output on success paths. The hardware-affecting findings (e.g. `toggle_powermode` ignoring the read error) are intentionally left out and will be reported separately for maintainer input.

## Changes

- **`sensor_show`: avoid multiplying an uninitialized value on the error path** (`legion-laptop.c:7396-7401`)
  `outval *= 1000` ran even when `read_temperature()` failed, multiplying stack garbage (UB). Now it only scales when the read succeeded. The subsequent `if (err) return err;` already discarded the bad value, so user-visible behavior is unchanged — this just removes the UB.

- **`debugfs_fancurve_show`: check `wmi_read_fancurve_custom()` error** (`legion-laptop.c:5440`)
  The previous code stored `err` but never checked it; on failure it printed a possibly-partial fan curve as if valid. On error it now prints the error line instead. Debug output only.

- **`fan_max_show` / `minifancurve_show`: use `sysfs_emit()` instead of `sprintf()`** (`7477`, `7951`)
  Same class as the previous `sprintf()` → `sysfs_emit()` fixes; these two were missed. Success-path output identical.

- **`minifancurve_show`: drop dead store** (`7946`)
  `err = -EIO;` was immediately superseded by `return -EIO;` and never read.

- **`acpi_init`: drop dead store** (`8243`)
  `acpi_path = get_model_acpi_path(_model, ACPI_PATH_WRITE_RAPIDCHARGE);` was always overwritten (when the `_STA` check runs) or never used (when skipped) before the first read.

## Verification

- Lint: `tests/test_kernel_checkpath.sh` (the CI gate) — **0 errors, 0 warnings**.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -5438,7 +5438,10 @@ static int debugfs_fancurve_show(struct seq_file *s, void *unused)
 	seq_puts(s, "Current fan curve in hardware (WMI; might be empty)\n");
 	wmi_fancurve.size = 0;
 	err = wmi_read_fancurve_custom(priv->conf, &wmi_fancurve);
-	fancurve_print_seqfile(&wmi_fancurve, s);
+	if (err)
+		seq_printf(s, "WMI fancurve error: %d\n", err);
+	else
+		fancurve_print_seqfile(&wmi_fancurve, s);
 	seq_puts(s, "=====================\n");
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


st0nie:
Thanks @armoox — merged. All five findings are real issues and the fixes are minimal and correct: error paths no longer clobber/scale stale values, `minifancurve` keeps the original EC error code instead of overwriting with `-EIO`, `sprintf` → `sysfs_emit` for sysfs show functions, and the dead `acpi_path` assignment in `acpi_init()` is gone (it was reassigned before every use). CI (build + checkpatch + DKMS) all green.

PR #545, opened 2026-09-08
fix(kernel-module): avoid uninitialized value in WMI bool read and debug dump

armoox:
## Summary

Two more clang static-analyzer findings from the continuing audit, both error-path only — success outputs are byte-identical.

### 1. `get_simple_wmi_attribute_bool()` — no garbage write-through on failure

```c
unsigned long int_val = *value;      // read uninitialized caller var
int err = get_simple_wmi_attribute(...);
*value = int_val;                    // write it back even on failure
```

`int_val` was initialized from `*value` (uninitialized at the call site) and then written back into `*value` regardless of the WMI read result. On failure the caller's output variable was clobbered with that garbage.

```c
unsigned long int_val = 0;
int err = get_simple_wmi_attribute(...);
if (!err)
    *value = int_val;
return err;
```

### 2. `seq_file_print_with_error()` — no uninitialized value in debug dump

`debugfs_fancurve_show()` prints many sensor/fan/powermode lines via this helper. When a read failed, it still printed the value line — emitting an uninitialized stack word into `/sys/kernel/debug/legion/fancurve`.

```c
seq_printf(s, "%s error: %ld\n", name, err);   // always
if (!err)
    seq_printf(s, "%s: %d\n", name, value);    // only on success
```

Provided by the same `mutex_lock(&priv->fancurve_mutex)` dump, this is debug output only; lockfancontroller/minifancurve lines already used this "error or value" pattern.

## Verification

- Lint gate: `tests/test_kernel_checkpath.sh` — **0 errors, 0 warnings**.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -3641,10 +3641,11 @@ static int get_simple_wmi_attribute_bool(struct legion_private *priv,
 					 u32 method_id, bool invert,
 					 unsigned long scale, bool *value)
 {
-	unsigned long int_val = *value;
+	unsigned long int_val = 0;
 	int err = get_simple_wmi_attribute(priv, guid, instance, method_id,
 					   invert, scale, &int_val);
-	*value = int_val;
+	if (!err)
+		*value = int_val;
 	return err;
 }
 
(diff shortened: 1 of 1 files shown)


st0nie:
Thanks @armoox — merged. Both fixes are error-path-only with byte-identical success output: `get_simple_wmi_attribute_bool()` no longer writes an uninitialized value back on WMI failure, and the fancurve debugfs dump now prints an error line instead of garbage when a read fails. Consistent with the pattern already used by the lockfancontroller/minifancurve lines. CI all green.

PR #553, opened 2026-09-10
fix(cli): make argcomplete optional instead of a hard dependency

armoox:
**Problem**: `legion_cli.py` imports `argcomplete` at module load, but the code comment says autocompletion should be added "only if package is installed" (#???). If `argcomplete` is missing, the *entire CLI* fails to start (every subcommand dies with ImportError), even though completion is only a convenience feature.

**Fix**:
- Wrap the `argcomplete` import in `try/except ImportError` and set the module to `None` when unavailable.
- Only call `argcomplete.autocomplete(parser)` when it actually imported, preserving current behavior everywhere else.

**Testing**: pylint 10/10, black clean, py_compile passes; smoke-tested both the present path (CLI help renders fully) and the fallback path (argcomplete blocked -> `legion_cli.argcomplete is None`, CLI still imports and runs).

### DIFF
--- a/python/legion_linux/legion_linux/legion_cli.py
+++ b/python/legion_linux/legion_linux/legion_cli.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python3
 # PYTHON_ARGCOMPLETE_OK
 # pylint: disable=wrong-import-order
-import argcomplete
+try:
+    import argcomplete
+except ImportError:
+    argcomplete = None
 import argparse
 import logging
 import sys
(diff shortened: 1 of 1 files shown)


armoox:
@st0nie Not sure about this PR, but if it's OK merge it, if its not feel free to close it, test it first then if not good close it.

armoox:
The `test-in-docker-container` check failed during the **Arch container build step**, before any of this PR's code is even pulled in:

```
#13 ==> ERROR: There is no secret key available to sign with.
#13 error: command failed to execute correctly
```

This is the known flaky `pacman-key`/packaging infra step of the CI workflow, unrelated to the diff (which is 6 lines in `legion_cli.py`). All other 4 checks passed. Could you re-run the failed job (or ask me to push a trivial bump to re-trigger)? Thanks.

PR #554, opened 2026-09-10
fix(gui): restore missing random header message

armoox:
Two adjacent string literals in `MainWindow.set_random_header_msg()` were missing their comma, so Python silently concatenated them into a single entry. The 5-message random rotation actually only rotated between 4 entries, and the concatenated one read awkwardly. Added the missing comma.

Verified the list now has 5 distinct entries. pylint 10/10, black clean, py_compile passes.


armoox:
lol same random mistake just fixed it

PR #560, opened 2026-09-10
fix(scripts): correct smartfan docs and report fan_control write failures

armoox:
Maintenance audit follow-up. What changed:
- `smartfan README`: state the actual 5-sample temperature average (code uses `temp_samples=5`, docs claimed 8) and drop the unsubstantiated keyboard-LED `color sync` feature claim — the smartfan scripts only write the power mode, never any LED
- `smartfan install.sh`: the daemon starts in `balanced` mode (matches the .service unit), not `quiet`
- `fan_control.sh`: detect failed sysfs writes (e.g. `-EBUSY` when full speed is requested before the power mode is set to Custom on `fanfullspeed_requires_custom_powermode` models) and exit non-zero instead of printing a misleading `enabled!` success message


armoox:
The `test-in-docker-container` failure is an infrastructure flake, not the diff: the `legion-arch` container fails at image build time (`RUN pacman -Syu --noconfirm`) with `ERROR: There is no secret key available to sign with` before any repo code is pulled. Same failure hit PR #553. Could you re-run the job? I can't trigger re-runs from the fork without admin rights.

Issue #534, opened 2026-09-05
[BUG]


chndsh:
cannot write fan speed in any modes, with at most the starting column getting set but not the rest of the columns. 
~~~
INFO:legion_linux.legion:Click on UI Lock fan controller, lock temperature sensors, and lock current fan speed element for LockFanController
INFO:legion_linux.legion:Feature LockFanController writing: 0
INFO:legion_linux.legion:Feature LockFanController reading
INFO:legion_linux.legion:Feature LockFanController reading: 0
INFO:legion_linux.legion:Trying to set minifancurve using fancurve profile to: False
INFO:legion_linux.legion:Setting minifancurve to: False
OSError: [Errno 95] Operation not supported

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ch1/LenovoLegionLinux/python/legion_linux/legion_linux/legion_gui.py", line 872, in on_write_fan_curve_to_hw
    self.model.write_fancurve_to_hw()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/ch1/LenovoLegionLinux/python/legion_linux/legion_linux/../legion_linux/legion.py", line 1723, in write_fancurve_to_hw
    self.fancurve_io.write_fan_curve(self.fan_curve, write_minifancurve)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ch1/LenovoLegionLinux/python/legion_linux/legion_linux/../legion_linux/legion.py", line 1048, in write_fan_curve
    self.set_acceleration(point_id, entry.acceleration)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ch1/LenovoLegionLinux/python/legion_linux/legion_linux/../legion_linux/legion.py", line 924, in set_acceleration
    self._write_file(file_path, value)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/ch1/LenovoLegionLinux/python/legion_linux/legion_linux/../legion_linux/legion.py", line 853, in _write_file
    with open(file_path, "w", encoding=DEFAULT_ENCODING) as filepointer:
[...]

Issue #236, opened 2024-06-25
Lenovo Legion 7i Gen 7 (16IAX7) can't set fan curve or read fan curve


LastPuppet:
**Problem Description**
Whenever I try to read fan curve or set a fan curve via the gui, it crashes with "OSError: [Errno 95] Operation not supported". Don't know why this happens.
Here is the log when trying to read fan curve:
[legion_gui_log.txt](https://github.com/user-attachments/files/15970849/legion_gui_log.txt)

**Model and Debug Info**

Distribution: Endeavour OS (a distribution of Arch) Release Gemini
Model name: Lenovo Legion 7i Gen 7 (16IAX7)
CPU model: 12th Gen Intel(R) Core(TM) i7-12800HX
GPU model: NVIDIA GeForce RTX 3070 Ti Laptop GPU
Keyboard backlight: RGB
Light in lid or logo: yes
Light at IO-Ports at back: yes

Output of `sudo dmidecode -t system`. Please remove Serial Number and UUID for privacy:
[dmidecodeoutput.txt](https://github.com/user-attachments/files/15970769/dmidecodeoutput.txt)

Output of `sudo dmidecode -t bios`:
[dmidecodebiosoutput.txt](https://github.com/user-attachments/files/15970779/dmidecodebiosoutput.txt)

Output of `sudo cat /sys/kernel/
[fancurveoutput.txt](https://github.com/user-attachments/files/15970782/fancurveoutput.txt)
debug/legion/fancurve`:

LastPuppet:
Oops, github didn't intergrate the cat output, here it is:
[fancurveoutput.txt](https://github.com/user-attachments/files/15970865/fancurveoutput.txt)

Issue #140, opened 2023-11-15
[BUG] Fancurve not showing and EC Chip ID not 8227


Bryntet:
**Problem Description**
EC Chip ID is 5507 (not 8227) and fancurve showing as 0 everywhere.

Has the bug occurred following steps described in the README?
Yes

What steps did you take?
Installed it using the AUR

What commands did you enter?
`sudo cat /sys/kernel/debug/legion/fancurve`

What was the (full) error you received and what command or action did lead to it?
```
EC Chip ID: 5507
EC Chip Version: 2a4
legion_laptop features: fancurve powermode platformprofile platformprofilenotify minifancurve
legion_laptop ec_readonly: 0
ACPI CFG error: 0
ACPI CFG: 2081289482
temperature access method: 5
CPU temperature error: 0
CPU temperature: 75
CPU temperature EC error: 0
CPU temperature EC: 0
CPU temperature ACPI error: 0
CPU temperature ACPI: 75
CPU temperature WMI error: 0
CPU temperature WMI: 0
CPU temperature WMI2 error: 0
CPU temperature WMI2: 0
CPU temperature WMI3 error: 0
CPU temperature WMI3: 75
GPU temperature error: 0
GPU temperature: 50
GPU temperature EC error: 0
GPU temperature EC: 0
GPU temperature ACPI error: 0
GPU temperature ACPI: 50
GPU temperature WMI error: 0
GPU temperature WMI: 0
GPU temperature WMI2 error: 0
GPU temperature WMI2: 0
GPU temperature WMI3 error: 0
GPU temperature WMI3: 50
fan speed access method: 5
1 fanspeed error: 0
1 fanspeed: 2500
1 fanspeed EC error: 0
1 fanspeed EC: 48
1 fanspeed ACPI error: 0
1 fanspeed ACPI: 2500
1 fanspeed WMI error: 0
1 fanspeed WMI: 0
1 fanspeed WMI2 error: 0
1 fanspeed WMI2: 0
1 fanspeed WMI3 error: 0
1 fanspeed WMI3: 2500
2 fanspeed error: 0
2 fanspeed: 2700
2 fanspeed EC error: 0
2 fanspeed EC: 0
2 fanspeed ACPI error: 0
2 fanspeed ACPI: 2700
2 fanspeed WMI error: 0
2 fanspeed WMI: 0
2 fanspeed WMI2 error: 0
2 fanspeed WMI2: 0
2 fanspeed WMI3 error: 0
[...]

PR #563, opened 2026-09-11
fix(kernel-module): mark model_lpcn fancurve as WMI speed-only (16ARX8/LPCN family)

daniel-baf:
## Problem

On the LPCN family (e.g. **Legion Pro 5 16ARX8 / 82WM**, BIOS LPCN65WW, EC 0x5507), `access_method_fancurve = ACCESS_METHOD_WMI3`, but the WMI method only supports the per-point **speed**:

- `wmi_write_fancurve_custom()` builds a 0x40 buffer where only `buffer[0x06..0x18]` (points 0-9 `speed1`) are filled — temperature thresholds, fan2 speed and accel/decel are never transmitted.
- `wmi_read_fancurve_custom()` `memset`s the curve and only fills `speed1` — everything else reads back as 0.

Consequences on model_lpcn today:

1. Writes to `pwm*_auto_pointN_temp`/`_temp_hyst`/`pwm2_*` are **accepted and silently never persisted** (debugfs readback always 0).
2. Accel/decel writes fail kernel validation with `-EOPNOTSUPP`, so `legion_cli fancurve-write-file-to-hw` and the GUI "Apply to HW" crash with `OSError: [Errno 95] Operation not supported` **mid-write**, leaving a partially applied curve.
3. The all-zeros curve in debugfs confuses users into thinking fan control is broken (#140, same machine; same errno pattern on other WMI models: #534, #535, #236).

## Fix

Set `.wmi_fancurve_speed_only = true` on `model_lpcn`, mirroring `model_n2cn` and the EC4 attribute filtering from #537. The hwmon fancurve group then exposes only `fan1_max` and `pwm1_auto_point{1-10}_pwm`; userspace (python lib probes attribute existence) auto-skips the hidden fields, and full fancurve writes complete successfully. The EC keeps using its own firmware temperature thresholds — same behavior as before, minus the misleading no-op attributes and the crash.

Also set `.has_extreme_powermode = false` on `model_lpcn`: on this EC, the firmware's `max-power` mode doesn't raise limits like on other extreme-powermode models — it cuts power **instantly** (hard power-off, no shutdown sequence). Rather than just documenting the risk, the kernel module no longer exposes `max-power` as a `platform_profile` choice on this model, so it can't be reached manually or via legiond auto-switching.

## Testing

On a Legion Pro 5 16ARX8 (82WM, BIOS LPCN65WW), Arch Linux, kernel 7.2.4-arch1-2, built from main @ e3b2116 + this patch:

- [x] `make` builds clean
- [x] `sudo make reloadmodule` loads; hwmon exposes exactly `fan1_max` + 10 `pwm1_auto_pointN_pwm` files (was 101 fancurve files incl. the no-op ones)
- [x] `sudo legion_cli fancurve-write-file-to-hw custom-ac.yaml` → "Successfully wrote fan curve from file to hardware" (previously crashed at `set_acceleration` with Errno 95)
- [x] Speeds land in the EC and fans track them (verified via `/sys/kernel/debug/legion/fancurve` and live RPM); EC thermal control unaffected
- [x] Sensors/fan RPM telemetry unaffected
- [x] `platform_profile_choices` no longer lists `max-power` on this model

## README

- Adds the 16ARX8 (82WM) to the confirmed models list (speed-only fan curve) with the LPCN note, refs #140.
- Documents the LPCN65WW `max-power` hard power-off quirk (and that it can reset EC battery charge mode from conservation/Long_Life back to rapid charge), and notes it's now hidden from `platform_profile` on this model rather than just a warning.
- The WMI3 speed-only note is scoped as confirmed on the Legion Pro 5 16ARX8 (LPCN65WW) and assumed (not yet confirmed) for the rest of the LPCN family, e.g. LPCN47WW.

Would appreciate confirmation from other LPCN users (e.g. Legion 7 Pro 16ARX8H, LPCN47WW) that speed-only matches what they see — based on the shared `wmi_write_fancurve_custom()` path it should apply to the whole family.

---

**Follow-up:** per AGENTS.md — checkpatch clean (0 errors, 0 warnings), C hunk clang-format clean, and README_zh-hans.md carries the same model entry and Known Bugs bullets (docs parity). Model naming matches the official Lenovo name ("Legion Pro 5"), consistent with the existing "Legion Pro 7" / "Legion 7 Pro" entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ It allows you to control features like the fan curve, power mode, power limits, 
 - Lenovo Legion 5 17ACH6 (BIOS HHCN31WW): sensors, fan curve, power profile
 - Lenovo Legion 7i 16ITHG6 (BIOS H1CN35WW): sensors, fan curve, power profile
 - Lenovo Legion 7 Pro 16ARX8H (BIOS LPCN47WW): sensors, fan curve, power profile
-- Lenovo Legion Pro 5 16ARX8 (82WM) (BIOS LPCN65WW): sensors, fan curve (speed points only, see LPCN note below), power profile
+- Lenovo Legion 5 Pro 16ARX8 (82WM) (BIOS LPCN65WW): sensors, fan curve (speed points only, see LPCN note below), power profile
 - Lenovo Legion 7 16IAX7 (82TD) (BIOS K1CN48WW): sensors, fan curve (write works; WMI readback returns empty buffer), power profile
 - Lenovo Legion Pro 7 16IRX8H (BIOS KWCN54WW): sensors, fan curve, power profile, fan unlock (lifts the fan ceiling from ~4400 to ~7100 RPM)
 - Lenovo Legion 7 16IRX9, Gen 9: sensors, fan curve, power profile; also supported by [SmartFan](extra/smartfan/) without the kernel module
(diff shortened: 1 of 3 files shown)


armoox:
@daniel-baf Thanks for the patch — the core change is good: it reuses the same wmi_fancurve_speed_only mechanism model_n2cn already uses, and the Python lib auto-skips hidden attributes, so the crash workaround holds.

Two things to fix before merge:

1. **Model name.** The 82WM is officially the Lenovo Legion Pro 5 16ARX8 (Lenovo PSREF/support pages, and the machine in #140 reports it as "Legion Pro 5 16ARX8"). The PR uses "Legion 5 Pro 16ARX8" in README.md, README_zh-hans.md and the description. Please rename to "Legion Pro 5" to match the existing "Legion 7 Pro 16ARX8H" entry.

2. **README states family-wide behavior as fact, but you asked for family confirmation.** The README bullets assert the speed-only limitation and the max-power power-cut for the whole LPCN family, yet you tested only LPCN65WW and noted you'd like confirmation from 16ARX8H (LPCN47WW) users. Consider wording those bullets as "confirmed on LPCN65WW, applies to the rest of the family by inference" until then.

Suggestion (not a blocker): the max-power power-cut is documented as an instant hard power-off, but has_extreme_powermode = true still exposes max-power in platform_profile (and legiond auto-switches profiles). Consider gating it off for LPCN65WW or the family if confirmed.

daniel-baf:
@armoox thanks for the review, addressed in 19690aa:

1. **Model name**: reverted back to the official "Legion Pro 5 16ARX8" in README.md and README_zh-hans.md (an earlier docs-alignment commit had accidentally flipped it to "Legion 5 Pro").
2. **Family-wide claim**: reworded the WMI3 speed-only bullet (both README and zh-hans) to state it's confirmed on the Legion Pro 5 16ARX8 (LPCN65WW) and assumed — not yet confirmed — for the rest of the LPCN family (e.g. LPCN47WW), pending confirmation from those users.
3. **max-power gating suggestion**: implemented it — set `has_extreme_powermode = false` for `model_lpcn`, so `max-power` is no longer exposed as a `platform_profile` choice on this model at all (can't be reached manually or via legiond auto-switching). Updated the README bullet to describe it as hidden rather than just a warning.

PR description updated to match.

Issue #564, opened 2026-09-12
[BUG] Legion Pro 5 16IRX8 (82WK, KWCN54WW): 0.0.26 does not probe on kernel 7.x; Fan_Set_Table takes level indices 0..10 (measured), stock reads silent zeros


Hishammm0:
**Problem Description**

Two problems on this machine, both fixed in the attached patches (PRs to follow, referencing this issue):

1. **Release 0.0.26 does not probe on kernel 7.x** (any model with `acpi_check_dev`). Following the README ("Quick Test: Module is properly loaded"), `make LLVM=1` (clang-built CachyOS kernel; the DKMS package fails there, see #358) and `sudo make reloadmodule` give:

   ```text
   legion legion: Using configuration for system: KWCN
   legion legion: No ACPI handle, will use FQN paths
   legion legion: Could not evaluate ACPI _STA
   legion legion: Could not init ACPI access: -5
   legion legion: probe with driver legion failed with error -5
   ```

   Since 18953e1 the driver binds a virtual platform device on 7.0+ (no ACPI companion), and c52b4f2 made `eval_int()` / `exec_simple_method()` resolve every ACPI name from the root there. The default paths (`_STA`, `_CFG`, `VPC0.GBMD`, …) are relative to the embedded controller device, so `\_STA` does not exist, and the EC path differs per platform anyway (here `\_SB_.PC00.LPCB.EC0_`). With that fixed, the fatal `_CFG` check fails next: on this DSDT the only `_CFG` is `VPC0._CFG` (ideapad layout), the EC has none, and `_CFG` is only printed, never used.

2. **`model_kwcn` treats the WMI3 fan table as percent; it is level indices 0..10.** `LENOVO_FAN_TABLE_DATA` publishes one nominal RPM per level (identical in every mode):

   | level | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
   |---|---|---|---|---|---|---|---|---|---|---|
   | fan 1 | 1700 | 1900 | 2100 | 2300 | 2500 | 2900 | 3400 | 3700 | 4400 | 5400 |
   | fan 2 | 1700 | 1900 | 2100 | 2200 | 2700 | 2900 | 3500 | 3700 | 4600 | 5400 |

[...]

PR #565, opened 2026-09-12
fix(kernel): probe on kernel 7.x - evaluate EC ACPI names relative to the PNP0C09 device; do not fail on a missing _CFG

Hishammm0:
Model: **Lenovo Legion Pro 5 16IRX8 (82WK), BIOS KWCN54WW.** The fix itself is not model-specific.

Part of #564.

On kernel 7.0+ the driver binds a virtual platform device (18953e1), so `priv->adev` is NULL, and c52b4f2 made
`eval_int()` / `exec_simple_method()` resolve every ACPI name from the root on those kernels. The default paths
(`_STA`, `_CFG`, `VPC0.GBMD`, `VPC0.SBMC`, `FANS`, `CPUT`, …) are names relative to the embedded controller device
the driver bound to before 7.0, so `\_STA` does not exist and the probe fails for **every model that sets
`acpi_check_dev`** (16 models rely on the relative defaults). The EC path also differs between platforms (here
`\_SB_.PC00.LPCB.EC0_`), so a single absolute default cannot fix it.

Release 0.0.26 unpatched, CachyOS 7.2.4-3-cachyos:

```text
legion legion: Using configuration for system: KWCN
legion legion: No ACPI handle, will use FQN paths
legion legion: Could not evaluate ACPI _STA
legion legion: Could not init ACPI access: -5
legion legion: probe with driver legion failed with error -5
```

**1. Evaluate the relative names against the embedded controller again.** On 7.x, look the EC up by HID
(`acpi_dev_get_first_match_dev("PNP0C09", NULL, -1)`, which only returns present devices) in `acpi_init()` and
evaluate against its handle, exactly as the ACPI companion provided before 7.0. Absolute per-model overrides keep
working because ACPICA ignores the handle for rooted names. The reference is dropped in a new `acpi_exit()` from
`legion_remove()` and from the probe error path; the root-resolution fallback stays for the case where no PNP0C09
device exists. `eval_int()` and `exec_simple_method()` lose their 7.x special case, because the existing `!adev`
branch already does the right thing.

**2. Do not fail the probe on a missing `_CFG`.** The value is only reported (probe log and debugfs) and never
gates a feature; on this DSDT the only `_CFG` is `VPC0._CFG` (ideapad layout) and the EC device has none. Log the
result either way and continue. `_STA` remains the presence check.

## Verification

`make` then `sudo make reloadmodule` on 7.2.4-3-cachyos, 0.0.26 plus these two commits:

```text
legion legion: Using ACPI device PNP0C09:00 for EC methods
legion legion: Could not evaluate ACPI _CFG: -5      (stock model_kwcn; the model paths in the 16IRX8 PR resolve it)
legion legion: Read embedded controller ID 0x5507
legion legion: legion_laptop loaded for this device
```

`sensors` then lists `legion_hwmon-isa-0000` with Fan 1 / Fan 2 / CPU / GPU / IC temperatures, and
`sudo cat /sys/kernel/debug/legion/fancurve` prints the ten-point table (README quick tests 1-3). Load and unload
cycles are clean, and the fan-curve write test in the 16IRX8 PR was run on top of this. Kernels older than 7.0 keep
the ACPI-companion path unchanged: the `acpi_dev_put()` is under the 7.0 guard.

`tests/test_kernel_checkpath.sh` reports 0 errors and 0 warnings, and the touched code is clang-format clean.

### DIFF
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -8314,17 +8314,24 @@ static int acpi_init(struct legion_private *priv, struct acpi_device *adev)
 		acpi_path = get_model_acpi_path(_model, ACPI_PATH_STA);
 		err = eval_int(priv->adev, acpi_path, &cfg);
 		if (err) {
-			dev_info(dev, "Could not evaluate ACPI _STA\n");
+			dev_info(dev, "Could not evaluate ACPI %s: %d\n",
+				 acpi_path, err);
 			goto err_acpi_init;
 		}
 
+		/*
+		 * _CFG is only reported (here and in debugfs), never used to
+		 * gate a feature, and not every Lenovo DSDT defines it on the
+		 * embedded controller device, so its absence must not stop
+		 * the probe; _STA above is the presence check.
+		 */
 		acpi_path = get_model_acpi_path(_model, ACPI_PATH_CFG);
 		err = eval_int(priv->adev, acpi_path, &cfg);
-		if (err) {
-			dev_info(dev, "Could not evaluate ACPI _CFG\n");
-			goto err_acpi_init;
-		}
-		dev_info(dev, "ACPI CFG: %lu\n", cfg);
+		if (err)
+			dev_info(dev, "Could not evaluate ACPI %s: %d\n",
+				 acpi_path, err);
+		else
+			dev_info(dev, "ACPI CFG: %lu\n", cfg);
 	} else {
 		dev_info(dev, "Skipping ACPI _STA check");
 	}
--- a/kernel_module/legion-laptop.c
+++ b/kernel_module/legion-laptop.c
@@ -2243,14 +2243,10 @@ static int eval_int(struct acpi_device *adev, const char *name,
 	unsigned long long result;
 	acpi_status status;
 	acpi_handle handle;
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(7, 0, 0)
-	status = acpi_get_handle(NULL, (char *)name, &handle);
-	if (ACPI_FAILURE(status))
-		return -EIO;
-#else
+
 	if (!adev) {
-		/* No ACPI companion device: resolve the full-qualified name
-		 * from the ACPI root, same as on kernel 7.0+.
+		/* No ACPI device to evaluate relative to: resolve the
+		 * fully-qualified name from the ACPI root.
 		 */
 		status = acpi_get_handle(NULL, (char *)name, &handle);
 		if (ACPI_FAILURE(status))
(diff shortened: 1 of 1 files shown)


armoox:

### Look good

- The unification of `eval_int()`/`exec_simple_method()` is behavior-identical on kernels < 7.0 (the `!adev` root-resolution fallback is preserved) and it fixes the >= 7.0 case, where the relative DSDT paths (`VPC0.GBMD`, `_STA`, `_CFG`) were previously resolved from the ACPI root instead of the EC device. That's the actual probe bug.
- `acpi_dev_get_first_match_dev("PNP0C09", …)` + `acpi_dev_put()` lifecycle is balanced: the reference is only ever taken on >= 7.0 where `acpi_init()` is handed `NULL`, and `acpi_exit()` puts exactly once (error path and `legion_remove`, mutually exclusive).

### Needs a maintainer decision: relaxing the `_CFG` gate affects every model, not just kernel 7.x

Making a missing/failing `_CFG` non-fatal changes probe behavior for **all models with `acpi_check_dev = true` on all kernels**, not just the kernel-7 case. Previously a failed `_CFG` evaluation aborted the probe (module refused to bind on that machine); after this change the driver binds as long as `_STA` passes. That loosens the effective allowlist validation for the whole family.

This is defensible — `_CFG` is only used for logging/debugfs, never to gate a feature — but it should be an explicit maintainer call (@qquique / @st0nie), because it widens which machines the module will load on. Consider calling it out in the PR description so that intent is on the record.

### Minor

- `acpi_dev_get_first_match_dev("PNP0C09", NULL, -1)` grabs the first PNP0C09 device; standard Legion platforms have a single EC, but if a DSDT ever declares more than one, the first match may be a battery/sensor EC. Worth a comment.
- On kernel < 7 the `#if` in `acpi_init` means `priv->adev` can be the caller's `ACPI_COMPANION` reference — `acpi_exit()` correctly guards the put, but the dual ownership (`priv->adev` is sometimes borrowed, sometimes owned) could use a one-line comment documenting when the driver owns the reference.

**Verdict:** the fix itself is sound; the `_CFG` probe relaxation needs explicit maintainer sign-off before merge.

st0nie:
Review feedback addressed and merged as 654ade4.

**Minor items** — both fixed in b141d99 (comments only, no behavior change):
- `acpi_dev_get_first_match_dev("PNP0C09", …)` first-match semantics are now documented, including the fallback to absolute per-model paths if a DSDT ever declares more than one PNP0C09 device.
- Dual ownership of `priv->adev` (borrowed ACPI-companion reference before 7.0, driver-owned reference on 7.0+) is now documented at the assignment, so the version-gated `acpi_dev_put()` in `acpi_exit()` is clear.

**_CFG probe relaxation** — maintainer sign-off given: the intent (value is only logged / shown in debugfs, never gates a feature; `_STA` remains the presence check) is already on record in the PR description, and the wider bind surface is acceptable since feature access is still gated per-model by the DMI allowlist and `acpi_check_dev` configs.

Thanks @armoox for the thorough review.

PR #566, opened 2026-09-12
fix(build): build with the LLVM toolchain automatically on clang-built kernels

Hishammm0:
Model: **Lenovo Legion Pro 5 16IRX8 (82WK), BIOS KWCN54WW**, on CachyOS 7.2.4-3-cachyos. The fix is not
model-specific; it applies to any kernel built with clang.

Fixes #358. Part of #564.

Out-of-tree builds against a kernel compiled with clang (`CONFIG_CC_IS_CLANG=y`, e.g. CachyOS) fail with gcc,
because the kernel's own flags are clang-only:

```text
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: clang version 22.1.8
  You are using:           gcc (GCC) 16.2.1
gcc: error: unrecognized command-line option '-mstack-alignment=8'
gcc: error: unrecognized command-line option '-mretpoline-external-thunk'
gcc: error: unrecognized command-line option '-mllvm'
```

Detect `CONFIG_CC_IS_CLANG` in the target kernel's configuration and pass `LLVM=1` to the kernel build unless
`LLVM` is given on the command line, so that plain `make`, `make install` and the DKMS build all work without a
manual flag. dkms 3.x does the same detection for the build command it runs itself, but `dkms.conf` here invokes
this Makefile, so the Makefile has to know too.

Three smaller fixes come with it, all in the same area:

- `LLVM` is always passed to the sub-make explicitly, with an empty value when it is 0. Without that, make carries
  the detected `LLVM=1` into the sub-make through `MAKEOVERRIDES` and a command-line `LLVM=0` has no effect.
- The detection reads `include/config/auto.conf` as well as `.config`, with `grep -qs` so a missing file is quiet.
- `dkms.conf` passes `KSRC=${kernel_source_dir}`, so `dkms --kernelsourcedir` and the detection cannot disagree,
  and `make install` / `make uninstall` run `depmod -a $(KERNELVERSION)` instead of the undefined `$(KVER)`, which
  had them updating the running kernel's module index when building for a different kernel.

## Verification

On CachyOS 7.2.4-3-cachyos (clang 22.1.8):

```text
$ make                     # was: gcc errors as above
make -C /lib/modules/7.2.4-3-cachyos/build M=.../kernel_module LLVM=1 modules
  CC [M]  legion-laptop.o
  LD [M]  legion-laptop.ko
$ make LLVM=0              # override honoured: gcc is selected again
  The kernel was built by: clang version 22.1.8
  You are using:           gcc
```

[...]

PR #509, opened 2026-08-20
Obtains Fan Table Default data from bios

qquique:
- new wmi api function : wmi_exec_querys_int
- uses WMI GUID 87FB2A6D-D802-48E7-9208-4576C5F5C8D8 to obtain fan table default data based on an indice
- .has_fancurve_defaults has to be set to true for model_config
- exposes fan table default data for current powermode in sysfs /sys/class/platform-profile/platform-profile-?/device/hwmon/hwmon?/

  `fantable_cpu_default_data` 
  `fantable_gpu_default_data` 
  `fantable_ic_default_data`

ex: 
```bash
# laptop in powermode balanced
$ cat device/hwmon/hwmon4/fantable_cpu_default_data
2,1,10,0,1400,1700,1900,2200,2500,2900,3100,3500,3900,4,10,70,73,77,80,84,84,92,95,99,100,0,10,3900,9,0,0,100,100,0,1
```

  Order of fields equals as struct WMIFanTableDefaultData

  note: "expose" doesnt mean the values are being used on the EC, is what it can be used when vantage app or this driver apply the custom curves.

- change with Fn+Q , /sys/class/platform-profile/platform-profile-?/ profile gets current powermode fan table default data.
- in Custom mode fancurve_defaults_powermode can be used to set the default values of the fancurve for some models as in PR #438,

So it will help for the users that has laptop models with default fancurves, to see something like :

From this:
```
Current fan curve in hardware:
Fan curve current point id: 4
Fan curve points size: 10
u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
3	 14	 14	 35	 35	 0	 0	 0	 60	 0	 60	 0	 42
3	 17	 17	 43	 43	 0	 0	 50	 64	 50	 64	 40	 63
3	 19	 19	 48	 48	 0	 0	 56	 68	 56	 68	 60	 72
3	 22	 22	 56	 56	 0	 0	 60	 72	 60	 72	 66	 74
3	 25	 25	 63	 63	 0	 0	 65	 76	 65	 76	 70	 76
3	 27	 27	 68	 68	 0	 0	 70	 80	 70	 80	 73	 78
3	 29	 29	 73	 73	 0	 0	 75	 84	 75	 84	 75	 80
[...]

PR #567, opened 2026-09-12
Legion Pro 5 16IRX8 (KWCN): WMI3 fan table as level indices, Other Method paths, RPM-to-level presets

### LINKED ISSUE #564: [BUG] Legion Pro 5 16IRX8 (82WK, KWCN54WW): 0.0.26 does not probe on kernel 7.x; Fan_Set_Table takes level indices 0..10 (measured), stock reads silent zeros

Hishammm0:
Model: **Lenovo Legion Pro 5 16IRX8 (82WK), BIOS KWCN54WW, EC 0x5507**, i9-13900HX + RTX 4070 Laptop, on CachyOS
7.2.4-3-cachyos.

Fixes #564 (the fan-table and Other-Method parts). Applies cleanly on `main` on its own, but on kernel 7.x the
driver needs the probe fixes in the companion PR to load at all.

Everything below was measured, first on Windows with the same WMI calls Lenovo Legion Toolkit makes, then on Linux
through this driver's hwmon interface.

## `Fan_Set_Table` takes level indices 0..10, not percentages

The firmware publishes one nominal RPM per level and fan (`LENOVO_FAN_TABLE_DATA.FanTable_Data`, identical in every
power mode):

| level | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| fan 1 | 1700 | 1900 | 2100 | 2300 | 2500 | 2900 | 3400 | 3700 | 4400 | 5400 |
| fan 2 | 1700 | 1900 | 2100 | 2200 | 2700 | 2900 | 3500 | 3700 | 4600 | 5400 |

In SmartFanMode 255, level 9 on every point gives **4400 / 4600 RPM** and level 10 gives **5400 / 5400 RPM**, the
table entries exactly. The fan 1 / fan 2 asymmetry at level 9 is the signature of a table lookup. The stock
`FAN_SPEED_UNIT_PERCENT` reads the firmware's default 1..10 back as pwm 2..25 and sends out-of-range levels on
write.

The 82WK DSDT confirms the mechanism. `\_SB_.GZFD.WMAB` method 6 creates FSTM@0, FSID@1, FSTL@2 and FSS0..FSS9,
never uses the first three, copies FSS0..FSS9 into `EC0.F9F0..F9F9` under the `LFCM` mutex, calls
`EC0.NCMD(0x8C, 0)` and returns 0 unconditionally. Method 5 ignores its argument, echoes the same ten bytes and
leaves the sensor half of the reply zero. So per-fan curves through FSID are unreachable here and the sensor table
is never populated.

`FAN_SPEED_UNIT_LEVEL` maps hwmon pwm 0..255 to level 0..10 and back, refuses a level below the per-point minimum
Legion Toolkit uses (1 for points 1-8, 3 and 5 for points 9-10) with `EOPNOTSUPP`, and sanitizes the whole table
before every transmit, because `Fan_Get_Table` returns ten zeros until the table has been written and an older
driver may have left percent values behind.

## The absent WMI methods do not fail, they return zeros

`LENOVO_FAN_METHOD` declares only `Fan_Get_Table` and `Fan_Set_Table` (no ids 1/2/3), `LENOVO_CPU_METHOD` only
`CPU_Set_OC_Data`, and `LENOVO_GPU_METHOD` nothing. On the stock driver `fan_fullspeed`, `fan_maxspeed` and six
power-limit attributes therefore read a silent **0** and three return EINVAL, because the WMI call's ACPI status is
never checked. `LENOVO_OTHER_METHOD.GetFeatureValue` answers the plain feature IDs and `LENOVO_CAPABILITY_DATA_01`
carries 70 per-mode rows with the ranges, which is exactly what `ACCESS_METHOD_WMI3_CLAMPED` consumes. So
`model_kwcn` moves to the Other Method paths, gains `wmi_fancurve_speed_only` (the temperature, pwm2 and
accel/decel attributes cannot be transmitted on this method) and `fan_max_rpm = 5400`.

## Userspace

`legion.py` converted RPM to pwm linearly through `fan1_max`, so on a level-index model a 4400 rpm preset became
level 8 (3700 rpm) and a 0 rpm point was refused by the driver. It now maps RPM to the nearest firmware level,
clamped to the point's minimum, and reads a level back as its nominal RPM. This covers `legion_cli`, the GUI and
`legiond`, which applies presets by running `legion_cli fancurve-write-preset-to-hw`.

## Verification

Balanced mode, after loading the patched module:

```text
$ D=/sys/bus/platform/drivers/legion/legion
$ cat $D/cpu_longterm_powerlimit $D/cpu_shortterm_powerlimit $D/gpu_ctgp_powerlimit $D/fan_fullspeed
70 119 55 0
$ H=$(grep -l legion_hwmon /sys/class/hwmon/hwmon*/name | xargs dirname); cat $H/fan1_max; ls $H | grep -c pwm
5400
10
```

The debugfs table shows `u` = 5 with levels 1..10 reading back as pwm 25..255. Flat-table test in custom mode
(`echo 255 > $D/powermode`), CPU idle:

| write | fans after 30 s | `Fan_Get_Table` / EC RAM 0x1F0 |
|---|---|---|
| level 9 on all ten points | **4400 / 4600 RPM** | `09` ×10, stable |
| level 10 on all ten points | **5400 / 5400 RPM** | `0a` ×10, stable |
| pwm 25 on point 10 (level 1, below its minimum 5) | unchanged | `Operation not supported`, dmesg names the attribute |
| restore 1..10, power mode back to balanced | spinning down | 1..10 |

The driver also coexists with the mainline `lenovo_wmi_*` drivers when loaded with `enable_platformprofile=0`: the
mainline drivers keep `platform_profile` and the power limits, this driver keeps the fan curve, and the mainline
`custom` profile turns out to be the same firmware mode as `powermode` 255. Both READMEs and
`doc/FEATURES_AND_TESTING.md` document that.

Gates: `tests/test_kernel_checkpath.sh` 0 errors / 0 warnings, clang-format clean, `tests/test_python.sh` 10.00/10,
`tests/test_python_cli.sh` passes, `black --check` clean.

Not included on purpose: surfacing the `Fan_Get_Table` sensor half as temperature breakpoints, since the DSDT shows
it is zero by construction on this firmware. `fan_unlock` was not exercised; the firmware's own maximum is 5400 RPM
on this model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ It allows you to control features like the fan curve, power mode, power limits, 
 - Lenovo Legion 7 Pro 16ARX8H (BIOS LPCN47WW): sensors, fan curve, power profile
 - Lenovo Legion 7 16IAX7 (82TD) (BIOS K1CN48WW): sensors, fan curve (write works; WMI readback returns empty buffer), power profile
 - Lenovo Legion Pro 7 16IRX8H (BIOS KWCN54WW): sensors, fan curve, power profile, fan unlock (lifts the fan ceiling from ~4400 to ~7100 RPM)
+- Lenovo Legion Pro 5 16IRX8 (82WK, BIOS KWCN54WW): sensors, fan curve (level indices 0-10, custom power mode), power profile, power limits and fan full speed through the Other Method WMI path; fan curve verified on Linux in custom power mode (level 9 → 4400/4600 RPM, level 10 → 5400/5400 RPM)
 - Lenovo Legion 7 16IRX9, Gen 9: sensors, fan curve, power profile; also supported by [SmartFan](extra/smartfan/) without the kernel module
 
 Many more models — including LOQ models and 2024/2025 Legions like the Legion 7 16IAX10 — are supported; see the DMI allowlist in [`kernel_module/legion-laptop.c`](kernel_module/legion-laptop.c) for the full list.
--- a/README.md
+++ b/README.md
@@ -203,6 +203,8 @@ sudo zypper install dkms openssl mokutil
 
 ```bash
 sudo pacman -S linux-headers base-devel lm_sensors git dmidecode python-pyqt6 python-yaml python-argcomplete python-darkdetect
+# Only for kernels built with clang (e.g. CachyOS); the Makefile then builds with LLVM=1 by itself
+sudo pacman -S clang llvm lld
 # Install the following for installation with DKMS
 sudo pacman -S dkms openssl mokutil
 ```
@@ -225,6 +227,8 @@ cd LenovoLegionLinux/kernel_module
 make
 sudo make reloadmodule
 ```
+
+*Note:* on kernels built with clang (`CONFIG_CC_IS_CLANG=y`, e.g. CachyOS) the Makefile detects this and passes `LLVM=1` to the kernel build by itself, for `make`, `make install` and DKMS alike; install `clang`, `llvm` and `lld` first.
 
 **For further instructions, problems, and tests see `Initial Usage Testing` below. Do them first before a permanent installation.**
 
@@ -465,7 +469,7 @@ Expected:
 
 - the controller might have loaded default values if you pressed Ctrl+Q(or FN+Q on certain devices) to change the power mode or waited too long; then try again
 - The entries in the fan curve are set to their values. The other values are not relevant (marked with XXXX)
-- On Legion Zone v3 firmware (e.g. Legion Pro 5 16IRX8, BIOS KWCN) the table holds fan levels 0-10 (`u` = 5): a written pwm is rounded to the nearest level (write 0, 26, 51, ... 255) and reads back as `level * 255 / 10` (0, 25, 51, 76, 102, 127, 153, 178, 204, 229, 255), points 9 and 10 need at least pwm 64 and 115, and a write below a point's minimum returns `Operation not supported`
+- On Legion Zone v3 firmware (e.g. Legion Pro 5 16IRX8, BIOS KWCN) the table holds fan levels 0-10 (`u` = 5): a written pwm is rounded to the nearest level (write 0, 26, 51, ... 255) and reads back as `level * 255 / 10` (0, 25, 51, 76, 102, 127, 153, 178, 204, 229, 255), points 9 and 10 need at least pwm 64 and 115, and a write below a point's minimum returns `Operation not supported`. The Python tools (`legion_cli`, `legion_gui` and the `legiond` presets) convert RPM values to the nearest firmware level on these models (for example 4400 rpm → level 9) and raise a 0 rpm point to the point's minimum level, so existing presets keep working
 
 ```
 u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
@@ -957,6 +961,15 @@ A graphical GNOME applet uses `power-profiles-daemon` to change the power mode u
 For KDE, there is the graphical tool `powerdevil`, which also uses `power-profiles-daemon` internally.
 
 If KDE only shows `balanced` and `performance` in `/sys/firmware/acpi/platform_profile_choices` but the Legion device (for example `/sys/devices/pci0000:00/0000:00:1f.0/PNP0C09:00/platform-profile/platform-profile-1/choices`) includes `quiet`, check whether `lenovo_wmi_gamezone` is also loaded. If both providers are active, the global profile choices are the intersection of both providers and quiet mode might disappear. In that case, unload or blacklist `lenovo_wmi_gamezone` and keep `legion_laptop` as the single provider for power mode.
+
+Alternatively keep both drivers and let each do what the other cannot: load `legion_laptop` with `enable_platformprofile=0`, so it registers no platform-profile provider and leaves the power mode (`platform_profile`) and the power limits (`/sys/class/firmware-attributes/lenovo-wmi-other-*`) to the mainline drivers. Its own `powermode` and power-limit attributes stay available under the platform device; what the mainline drivers do not offer at all, and what you keep `legion_laptop` for, is the fan curve, the fan speeds and fan full speed. Put
+
+```text
+options legion_laptop enable_platformprofile=0
+softdep legion_laptop pre: ideapad_laptop lenovo_wmi_gamezone lenovo_wmi_other lenovo_wmi_events lenovo_wmi_capdata
+```
+
+into `/etc/modprobe.d/legion_laptop.conf`; the `softdep` line loads the mainline drivers first so they bind the WMI devices both drivers list. The mainline `custom` profile is the same firmware mode as `powermode` 255; select it through `/sys/class/platform-profile/platform-profile-N/profile` of the device named `lenovo-wmi-gamezone`, because the legacy `/sys/firmware/acpi/platform_profile` file refuses `custom` by design. Verified on a Legion Pro 5 16IRX8 with kernel 7.2.
 
 ### It almost works, but (some) temperature sensor/changing point in fan control or (some) fan speed is not working. What should I do?
 
--- a/README.md
+++ b/README.md
@@ -821,6 +821,7 @@ Thank you for your contribution for the Linux support:
 * [normaneye](https://github.com/normaneye), fixing GPU temperature bug in GUI
 * [Petingoso](https://github.com/Petingoso), fix script to run withou sudo
 * [XenHat](https://github.com/XenHat), fix README
+* [Hishammm0](https://github.com/Hishammm0), Legion Pro 5 16IRX8 (KWCN) fan levels, kernel 7.x ACPI probe fix, clang build support
 
 Also please tell me if it works or does not work on your laptop.
 
--- a/README.md
+++ b/README.md
@@ -374,9 +374,9 @@ u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|decele
 The fan curve is displayed as a table with the following columns:
 
 ```text
-u(speed_of_unit): unit for the speed (1- Percentage, 2-PWM, 3-RPM)
-speed1[u]: speed in rpm divided by 100 for fan1 at this point
-speed2[u]: speed in rpm divided by 100 for fan2 at this point
+u(speed_of_unit): unit for the speed (1-Percentage, 2-PWM, 3-RPM/100, 4-Percentage rounded to nearest, 5-Level index 0-10 into the firmware fan table)
+speed1[u]: speed of fan1 at this point, in the unit given by u
+speed2[u]: speed of fan2 at this point, in the unit given by u
 speed1[pwm]: speed in pwm (0-255) for fan1 at this point
 speed2[pwm]: speed in pwm (0-255) for fan2 at this point
 acceleration: acceleration time (higher = slower)
@@ -464,6 +464,7 @@ Expected:
 
 - the controller might have loaded default values if you pressed Ctrl+Q(or FN+Q on certain devices) to change the power mode or waited too long; then try again
 - The entries in the fan curve are set to their values. The other values are not relevant (marked with XXXX)
+- On Legion Zone v3 firmware (e.g. Legion Pro 5 16IRX8, BIOS KWCN) the table holds fan levels 0-10 (`u` = 5): a written pwm is rounded to the nearest level (write 0, 26, 51, ... 255) and reads back as `level * 255 / 10` (0, 25, 51, 76, 102, 127, 153, 178, 204, 229, 255), points 9 and 10 need at least pwm 64 and 115, and a write below a point's minimum returns `Operation not supported`
 
 ```
 u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
--- a/README_zh-hans.md
+++ b/README_zh-hans.md
@@ -384,9 +384,9 @@ u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|decele
 风扇曲线以表格形式展示，列说明如下：
 
 ```text
-u(speed_of_unit): 风扇速度的单位（1-百分比, 2-PWM, 3-RPM）
-speed1[u]: fan1 在该点的速度（rpm 除以 100）
-speed2[u]: fan2 在该点的速度（rpm 除以 100）
+u(speed_of_unit): 风扇速度的单位（1-百分比, 2-PWM, 3-RPM/100, 4-百分比、四舍五入, 5-固件风扇表等级索引 0-10）
+speed1[u]: fan1 在该点的速度（单位由 u 决定）
+speed2[u]: fan2 在该点的速度（单位由 u 决定）
 speed1[pwm]: fan1 在该点的 pwm（0-255）
 speed2[pwm]: fan2 在该点的 pwm（0-255）
 acceleration: 加速时间（数值越大越慢）
@@ -476,6 +476,7 @@ cat /sys/kernel/debug/legion/fancurve
 
 - 如果你按下 Ctrl+Q（或某些设备的 FN+Q）更改电源模式，或等待时间过长，控制器可能会加载默认值；此时请重试
 - 风扇曲线中的对应条目值应被设置为你输入的数值，其它值不相关（用 XXXX 表示）
+- 在 Legion Zone v3 固件上（例如 Legion Pro 5 16IRX8，BIOS KWCN），风扇表保存的是 0-10 的风扇等级（`u` = 5）：写入的 pwm 会四舍五入到最近的等级（写入 0、26、51 … 255），读回时为 `level * 255 / 10`（0、25、51、76、102、127、153、178、204、229、255）；第 9、10 点至少需要 pwm 64 和 115，低于某点最小值的写入会返回 `Operation not supported`
 
 ```
 u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
--- a/README_zh-hans.md
+++ b/README_zh-hans.md
@@ -213,6 +213,8 @@ sudo zypper install dkms openssl mokutil
 
 ```bash
 sudo pacman -S linux-headers base-devel lm_sensors git dmidecode python-pyqt6 python-yaml python-argcomplete python-darkdetect
+# 仅用于 clang 编译的内核（例如 CachyOS）；Makefile 会自动以 LLVM=1 构建
+sudo pacman -S clang llvm lld
 # 如需通过 DKMS 安装，请安装以下包
 sudo pacman -S dkms openssl mokutil
 ```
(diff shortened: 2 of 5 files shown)


armoox:

### Blocking: this conflicts with #568 — same feature, two designs

Both this PR and #568 add fan-level fan tables (`FAN_SPEED_UNIT_LEVEL = 5`, a `case FAN_SPEED_UNIT_LEVEL:` in `fancurve_set_speed_pwm()`/`fancurve_get_speed_pwm()`, a switch in `wmi_read_fancurve_custom()`, and the same README/doc paragraphs). Merging both will not compile (duplicate case labels). Please coordinate with the #568 author and get a maintainer decision on which design is canonical — this one uses a global `MAX_FAN_LEVEL` + per-point minimum table rejecting below-minimum writes, #568 uses per-model `wmi_fancurve_max_level`/`min_level` with clamp-on-write.

### Behavior change for the whole KWCN family — needs a split or maintainer sign-off

`model_kwcn` is **shared** by the Legion Pro 5 16IRX8 **and** the Legion Pro 7 16IRX8H (both BIOS KWCN54WW, EC 0x5507). This PR flips for the entire family, on the strength of verification on one Legion Pro 5 16IRX8:

- `access_method_fanfullspeed`: WMI → WMI3 + `fanfullspeed_requires_custom_powermode`
- `access_method_powerlimits`: → `ACCESS_METHOD_WMI3_CLAMPED`
- `wmi_fancurve_speed_only = true` (hides fan2/temperature/accel curve attrs)

The Pro 7 16IRX8H is a different chassis and already shows different fan-table behavior (issue #429: EC resets the table and ignores writes in mode 0xE0). Changing its `fanfullspeed` path, powerlimit clamping and hiding its curve attributes without 16IRX8H-user validation is a regression risk for existing supported hardware. Please either split the configs (keyed on a distinguishing DMI string, e.g. product name) or get explicit confirmation from a 16IRX8H user before this goes in.

Same applies to the Python `LEVEL_FAN_TABLES` keyed on the `KWCN` prefix in `legion.py` — it now applies the RPM↔level mapping to both machines.

### Minor

- `fancurve_level_min` per-point minima (1,1,...,3,5) are claimed to mirror Lenovo Legion Toolkit (`GetMinimumFanTableAsync`). That's reverse-engineered data with no firmware reference backing — fine, but it should be stated as an assumption and confirmed against the DSDT/EC before it gates writes family-wide.
- In the read path, a stored level above `MAX_FAN_LEVEL` only warns (`pr_warn_once`) and is then written back on the next write, where the sanitizer clamps it — acceptable, just noting the flow.

### What looks good

- The `wmi_write_fancurve_custom()` refactor to the `buffer[0x06 + 2*point]` loop is byte-for-byte equivalent to the current hardcoded offsets (0x06,0x08,…0x18) — the i==0..9 expansion matches exactly.
- The sanitize step (`fancurve_level_table_sanitize`) handles both the never-written all-zero table (default 1..10) and stale percent values from older drivers — good safety thinking.

**Verdict:** needs (a) conflict resolution with #568, (b) a per-model split or explicit maintainer + 16IRX8H-user sign-off for the family-wide behavior changes.

qquique:
WQA3 method is behind the call of WMI GUID 87FB2A6D-D802-48E7-9208-4576C5F5C8D8 , could be other WQA<?> for other laptop models, instead of using fixed values the fantable data should be cached from the wmi guid, then filtered by the field mode = current powermode, and then filtered by sensor id to obtain the rpm/temps for cpu (0x04) and gpu (0x05), then you can obtain number of levels (field fan_table_len), and MAX_RPM (field current_fan_max_speed). see #509. the indices correspond to the rpm/temp columns of the fantable data.

st0nie:
@qquique your review suggestion is now implemented in #571 (merged as d7d14a08): the kernel module caches the `LENOVO_FAN_TABLE_DATA` rows from the WMI block (GUID 87FB2A6D-D802-48E7-9208-4576C5F5C8D8, queried by row index like the DSDT's `GZFD.WQA3`/`WQA7`), filters by mode and sensor id (fan 1 / 0x04, fan 2 / 0x05), and exposes the per-level RPM ladders as `fan1_level_rpm_table`/`fan2_level_rpm_table` on the platform device. The Python tools read those at runtime — the hardcoded `LEVEL_FAN_TABLES` (and its BIOS-prefix lookup) is gone, so a new level-index model needs no userspace table at all.

One follow-up it enables: `fan_max_rpm` could now come from the row's `current_fan_max_speed` field instead of the model config. If you have a KWCN/Q7CN unit handy, `cat /sys/bus/platform/drivers/legion/legion/fan1_level_rpm_table` printing the known ladder (fan 1: 1700..5400 on KWCN) would be a great real-hardware confirmation of the row decode.

Hishammm0:
Hardware confirmation for #571 on the **Legion Pro 5 16IRX8 (82WK, BIOS KWCN54WW, EC 0x5507)**, CachyOS
7.2.4-3-cachyos, DKMS install of `main` (2f26060) plus #577 (which only touches the cached power mode; the table
decode is #571 as merged):

```text
$ cat /sys/bus/platform/drivers/legion/legion/fan1_level_rpm_table
1700 1900 2100 2300 2500 2900 3400 3700 4400 5400
$ cat /sys/bus/platform/drivers/legion/legion/fan2_level_rpm_table
1700 1900 2100 2200 2700 2900 3500 3700 4600 5400
$ cat /sys/class/hwmon/hwmon4/fan1_max
5400
```

Both ladders decode from `LENOVO_FAN_TABLE_DATA` to exactly the values I measured by hand for #567 (fan 1 and fan 2
respectively), and `fan1_max` now comes from the table.

Same build, end to end, in the README coexistence setup (mainline `lenovo-wmi-*` loaded, `legion_laptop
enable_platformprofile=0`): switching the mainline profile to `custom` makes legion `powermode` read 255; writing
level 9 to all ten points (pwm readback 229) ramps the fans 1900 → 2500 → 3100 → 3800/3700 → **4400 / 4600 RPM at
+30 s**, the EC mirror at F9F0..F9F9 shows `09` ×10, the mainline power-limit attributes are untouched, and the
restore (default ladder, profile `balanced`, `powermode` 2) is clean. `dkms build` also works on the clang kernel
with the auto-detection from #566, so the DKMS path is what I run daily now. Happy to post the full log if useful.

Issue #409, opened 2026-03-18
[Device Support] Lenovo Legion 7i 16IAX10 Gen 10 (83KY) - Core Ultra 9 275HX / RTX 5070 - Full EC Dumps (Power, KB, Fans)


Walava42:
**System Information:**
-  **Model:** Lenovo Legion 7i Gen 10 (16IAX10H)
-  **Motherboard ID:** 83KY
-  **BIOS Version:** RXCN27WW
-  **CPU:** Intel Core Ultra 9 275HX (Arrow Lake)
-  **GPU:** NVIDIA GeForce RTX 5070 Laptop
-  **OS/Kernel:** Fedora 43 - 6.19.8-200.fc43.x86_64

**The Problem:**
CPU power limit is hard-locked at 31W on Linux. Additionally, keyboard lighting modes 5 & 6 (Music reactive) are not functioning correctly.

**Attached Evidence (Full Diagnostic Package):**
I've performed extensive EC RAM probing on Windows to assist in mapping the 83KY board:
1. **Power Profiles:** EC dumps for Quiet (31W), Balanced, and Performance (115W+).
2. **Keyboard Lighting:** Individual dumps for all 6 hardware modes (including the non-working music modes 5 & 6).
3. **Fans:** Baseline vs. Max Stress (3DMark) dumps to help identify fan duty cycle registers.
4. **ACPI:** Full 'acpi_dump' from Linux.

**Developer Notes / Hex Diff Findings:**
Power Modes: Toggling profiles seems to alter offset 0x06.
Fan RPM Sensors: High fluctuation observed in the 0xB0 - 0xB7 range comparing Idle vs. 3DMark Max Stress dumps.

**Files:**

[Legion7i_83KY_EC_ACPI_Dumps.zip](https://github.com/user-attachments/files/26093086/Legion7i_83KY_EC_ACPI_Dumps.zip)

NamelessHoodie:
I have nearly the same model @Walava42 I might be willing/able to help with this.
The current functionality of 'legion_laptop' on my device... Well less is working than isn't. It is not natively compatible and I had to set the flag(I forget the name) which allows legiond to run at all... Worked to switch between power modes using 'FN+Q'


```
lenovolegionlinux-dkms/stable,now 0.0.20+ds-1 amd64 [installed]
lenovolegionlinux-sound/stable,now 0.0.20+ds-1 amd64 [installed]
[...]

PR #396, opened 2026-02-11
Add support for Legion Pro 7 16IAX10H (Q7CN, EC 0x5508)

ChaoticSi1ence:
Add model_q7cn config for the 2025 Legion Pro 7 16IAX10H (83F5):
- EC chip: ITE IT5508 (0x5508), verified on hardware
- BIOS code: Q7CN
- Fan curve/speed/temperature via WMI3
- Power mode and fan full speed via WMI
- RGB keyboard is USB-controlled (ACCESS_METHOD_NO_ACCESS)
- Based on model_nscn (Legion 7 16IAX9) as closest match

Also fix minor line-length formatting in platform_profile_probe and platform_profile_init to comply with 80-column kernel style.

PR #568, opened 2026-09-12
feat(kernel): add Legion Pro 7 16IAX10H (83F5, Q7CN) model config (on top of #567)

ChaoticSi1ence:
Adds the **Legion Pro 7 16IAX10H (83F5, BIOS Q7CN)** to the DMI allowlist. Model enablement only.

**Stacked on #567** — the first six commits here are #567's (`Hishammm0:feat/legion-pro-5-16irx8`) and drop out on rebase once it merges. This PR no longer carries its own fan-level implementation: `model_q7cn` opts into #567's `FAN_SPEED_UNIT_LEVEL` (one line in `wmi_read_fancurve_custom()`), and `LEVEL_FAN_TABLES` gains the Q7CN RPM ladder so RPM presets round to the right level. Own diff vs #567: 2 commits, +156 lines (model + DMI entry + `skip_lockfancontroller` flag, python table entry, docs).

Supersedes the withdrawn Q7CN attempts #379 / #396 / #403 / #407. The EC RAM base (`0xFE500400`, DSDT `ERAX`) is the one @nuclearcat found in #407; earlier proposals carried `0xFE00D400` / `0xFE0B0400` over from other generations, which explains the frozen `ecmemoryram` dumps on 83F5 (#385, #409, #333).

### Hardware validation (this unit)

Legion Pro 7 16IAX10H, product `83F5`, BIOS `Q7CN78WW`, EC 0x5508 fw 1.78, kernel 7.2.4 (CachyOS, clang-built, `make LLVM=1`), in-tree `lenovo_wmi_gamezone/other/capdata/events` and `ideapad_laptop` loaded alongside; module loaded with `ec_readonly=1 enable_platformprofile=0`, no `force`:

```
legion legion: Using configuration for system: Q7CN
legion legion: Loaded 70 capdata01 rows and 27 discrete data entries (4 features) from WMI capability data
legion_laptop: Successfully mapped embedded controller: 0xfe500400 (in RAM)/0x0 (in EC)
legion legion: Read embedded controller ID 0x5508
```

- `debugfs/legion/ecmemory` (Super-I/O port I/O at 0xC400), `ecmemoryram` (ioremap 0xFE500400) and the ACPI EC I/O space (`ec_sys`) are byte-identical apart from live sensor bytes.
- `Fan_Get_Table` read `1,2,3,4,5,6,7,8,8,8` in performance mode, the same bytes as EC `F9F0..F9F9` (`ecmemoryram+0x180`).
- `temp1/temp2` and `fan1/fan2` via WMAE match the in-tree `lenovo_wmi_other` hwmon exactly.
- `powermode` 3 → 2 → 3: `lenovo_wmi_gamezone`'s `platform_profile` and the EC `SPMO` byte (`0x11 → 0x10 → 0x11`) followed each write.
- Custom mode via `/sys/class/platform-profile/platform-profile-0/profile` → EC `SPMO = 0x13`; a no-op table write left the bytes unchanged; `pwm1_auto_point10_pwm = 230` set EC byte 10 to `09` (level 9), `204` restored `08`; leaving custom restored `SPMO = 0x11`. `rmmod` clean, `lenovo_wmi_gamezone` kept the GameZone GUID throughout.

Not exercised: `fan_fullspeed` (WMAE `0x04020000`; DSDT sets `FNST` unconditionally, reported to take effect only in custom mode on 83F5) and the rapid-charge attributes (`VPC0.GBMD/SBMC`, present in the DSDT).

### Firmware facts behind the config (this unit's DSDT)

- `WMAB` implements only `Fan_Get_Table(5)` / `Fan_Set_Table(6)`; set copies the low byte of the ten u16 slots at `0x06..0x18` to EC `F9F0..F9F9` and issues `LECR(0xD0,1,1,2)`, no range check, mode byte ignored. Values are level indices — `LENOVO_FAN_TABLE_DATA` (WQA3): level 1..10 = 1600..5200 RPM (fan 1), 1700..5400 (fan 2), 2300..6500 (fan 4); one table for all fans, temperature axis fixed by the EC; extreme mode returns a static `1..10` placeholder. The EC applies the table in custom mode on AC only.
- Fan RPM, temperatures, full speed and power limits go through the Other Method GUID (`WMAE`); the CPU Method GUID (`WMAC`) is an empty stub → `ACCESS_METHOD_WMI3_CLAMPED` for the limit attributes left visible (`skip_oc_controls` hides PL/OC; the in-tree firmware-attributes cover those).
- GameZone fan/temperature/keyboard/fan-unlock methods are absent; keyboard/lid lighting is a USB-HID ITE `048d:c197` device → `ACCESS_METHOD_NO_ACCESS`, Y-logo / IO-port lights skipped (cf. #402).
- `minifancurve` (0xC536) and `lockfancontroller` (0xC4AB) bytes are not declared on this EC → hidden (`has_minifancurve = false`, new `skip_lockfancontroller`, default off elsewhere).

https://claude.ai/code/session_01GbRSuM2vYBmciDPXcFKJTv


armoox:

### Blocking: this conflicts with #567 — they implement the same feature twice

Both this PR and #567 introduce fan-level fan tables:

- Both add `FAN_SPEED_UNIT_LEVEL = 5` to `enum fan_speed_unit`.
- Both add a `case FAN_SPEED_UNIT_LEVEL:` inside `fancurve_set_speed_pwm()` **and** `fancurve_get_speed_pwm()`.
- Both modify `wmi_read_fancurve_custom()` to select the unit.
- Both rewrite the same README / FEATURES_AND_TESTING fan-curve paragraphs.

If both merge as-is, the duplicate `case` labels will not compile. They are also two **different designs**: this PR uses per-model `wmi_fancurve_max_level`/`min_level` fields with clamp-on-write, while #567 uses a global `MAX_FAN_LEVEL` + a hardcoded per-point minimum table and rejects below-minimum writes. Upstream can merge only one; the two authors need to coordinate (pick one design, rebase the other), and a maintainer (@qquique / @st0nie) should pick between the two.

### Minor

1. `legion_wmi_fancurve_speed_attribute()` gains `fancurve_defaults_powermode`. Today the only `wmi_fancurve_speed_only` model is `model_n2cn`, which does not set `has_fancurve_defaults` — and this PR's own `model_q7cn` sets `has_fancurve_defaults = false`. So this line is currently dead for every existing model. Either it's intended for a follow-up (fine, but say so) or it should be dropped until needed.

2. The comment on `skip_lockfancontroller` says it hides the attribute on models whose EC does not declare that byte — other models with the same condition (LOQ generation) may need it too; fine as a follow-up, just noting.

### What looks good

- DMI entry is product-qualified (`83F5` + BIOS `Q7CN`), following the repo rule of not widening beyond validated model/BIOS combos.
- The DSDT line references in the `model_q7cn` comment block are genuinely useful for review.
- Read-side clamping of out-of-range levels with `pr_warn_ratelimited` is a good defensive choice.

**Verdict:** durable work, but it needs (a) resolution of the conflict with #567 and (b) a maintainer decision on the fan-level design before it can be considered for main.

ChaoticSi1ence:
Thanks for the review — agreed that two level-unit designs can't both land. I've dropped mine rather than compete: this PR is now **stacked on #567** and reduced to model enablement only.

- Rebased onto `Hishammm0:feat/legion-pro-5-16irx8`; the first six commits are #567's and disappear on rebase once it merges.
- `model_q7cn` opts into #567's `FAN_SPEED_UNIT_LEVEL` via the existing pointer check in `wmi_read_fancurve_custom()`; no new `wmi_fancurve_max_level`/`min_level` fields, no duplicate enum/case labels, no rewrite of the shared README/FEATURES paragraphs.
- `LEVEL_FAN_TABLES["Q7CN"]` adds this firmware's RPM ladder (from the DSDT's `LENOVO_FAN_TABLE_DATA`) so #567's RPM→level preset mapping works here too.
- The `fancurve_defaults_powermode` allow-list line is gone with the dropped commit.
- `skip_lockfancontroller` kept as a plain opt-in flag (default off); happy to fold it into a follow-up for the LOQ generation if that's preferred.

Own diff vs #567 is 2 commits / +156 lines (model + DMI entry + flag, python table entry, docs). Hardware evidence is unchanged and in the description.

Issue #570, opened 2026-09-12
Fan-level WMI fan tables: PRs #567/#568 conflict needs a maintainer decision


armoox:
**Fan-level WMI fan tables: PRs #567 and #568 conflict — one design needs to win.**

The two open fan-curve PRs are competing implementations of the same feature (fan-level WMI fan tables) and cannot both land:

- Both add `FAN_SPEED_UNIT_LEVEL = 5` to `enum fan_speed_unit` and a duplicated `case FAN_SPEED_UNIT_LEVEL:` inside `fancurve_set_speed_pwm()` / `fancurve_get_speed_pwm()` → duplicate case labels once both are applied → build break.
- Both rework `wmi_read_fancurve_custom()` and the same README / FEATURES_AND_TESTING fan-curve sections → merge conflicts.
- Different designs:
  - **#567** (KWCN): global `MAX_FAN_LEVEL` + hardcoded per-point minimum table (rejects below-minimum writes, mirroring Lenovo Legion Toolkit).
  - **#568** (Q7CN): per-model `wmi_fancurve_max_level` / `min_level` with clamp-on-write.

Suggested resolution for a maintainer (@qquique / @st0nie): pick one canonical design, let the other author rebase onto it, and keep the per-model max/min data where it belongs (model config) so future models don't need new enum values.

Related: both PRs also carry family-wide `model_config` changes that this repo's rules say to avoid without per-model validation (details in the review comments on #567, #568, #563, #565). These should be handled as separate changes from the fan-level design so they don't ride along with it.

(Previously posted as a comment on #485; moved here since this is a separate concern from capdata validation.)

qquique:
I'm not maintainer, but I would prefer : 
- a PR that handle the levels/indices in its own sysfs, something like this commit [238289](https://github.com/johnfanv2/LenovoLegionLinux/pull/347/changes/23828931a63b20e8a952ab6dfca9a5d260c6dc71)
- another PR that modifies the client to use the indices/levels
- another PR that finishes PR #509 , and modifies the client to use the indices/levels to show the RPM/Temp values from the fantable.

st0nie:
Resolved — the conflict this issue tracks no longer exists:

[...]

Issue #485, opened 2026-08-07
[BUG] The writing of values using wmi_common_method_other_store for the power limits should need another step using another WMI calls


st0nie:
> Something to take in consideration : 
> 
> 
> 
> The writing of values using `wmi_common_method_other_store` for the power limits should need another step using another WMI calls (and caching) for the capability data of the laptop models, that capability data gives the constraint of what values can be modified and if it is in a range of values, because an out of range/incorrect value could damage the HW. That capability data currently on latest kernel is exposed via lenovo-wmi-capdata. To access that WMI info this driver would need to be modified for latest kernel and implemented/duplicate functionality for older kernels. 

 _Originally posted by @qquique in [2b5d6b4](https://github.com/johnfanv2/LenovoLegionLinux/commit/2b5d6b4c9a3f3ea2702c38c9b1040c8d5d0a8e67#commitcomment-195342253)_

st0nie:
Addressed: the mechanism qquique described exists as `ACCESS_METHOD_WMI3_CLAMPED` — `clamped_value()` looks up the LENOVO_CAPABILITY_DATA_01 row for the feature + current power mode (plus LENOVO_DISCRETE_DATA for discrete ranges) and clamps/snaps the value before `wmi_common_method_other_store` sends it. When no capdata row exists for the feature it passes through unchanged, so clamping degrades gracefully.

It is enabled per validated model — currently model_kwcn, model_secn and model_q7cn (and the peak/APU-SPPT attributes route through it as of a9e8fd2). We intentionally don't blanket-switch every WMI3 model without per-model validation (same policy as the fan-table work in #491). If a specific model writes out-of-range values, name it and we'll move that model to CLAMPED after checking its capdata rows. Closing as implemented.

PR #347, opened 2025-08-07
Fixes LOQ model fancurves

qquique:
Hi me again :), tldr of this PR :
- Fix for fancurves for LOQ Model (Tested on LZCN model),  reads and sets values when in Custom Mode for CPU Fan (min temp, max temp, rpm), GPU Fan (min temp, max temp, rpm), IC (min temp, max temp), and works!.
- WMI methods to read/write current Fan Indices (read details)

Now the details :

When using the Lenovo Vantage Application when setting the custom mode and modifying its levels (points) :
* Only changes position indices in an array that has the same size of the Fancurve.
* Each element value of that array points to the element that contains the speed in the current Fancurve selected.
* The Fancurve is selected depending on the thermal mode (quiet/balanced/performance/custom), the dGPU (by vendor id) and a PRID (value in DSDT) that I think is the mode of the laptop (hybrid, iGPU, dGPU, and another that wakes up the dGPU when used according to the Vantage app)

It can be queried with the following powershell script : 
```powershell
$classInstance = Get-CimInstance -Namespace root/wmi -ClassName LENOVO_FAN_METHOD
$classInstance | Invoke-CimMethod -MethodName Fan_Get_Table -Arguments @{FanID=0;SensorID=0}
```

```txt
FanTable        : {2, 2, 3, 4, 5, 6, 7, 8 ,9 ,9}
FanTableSize    : 10
SensorTable     : {2, 2, 3, 4, 5, 6 ,7, 8, 9 ,9}
SensorTableSize : 10
ReturnValue     : True
```
In this output the value 2 is repeated because I changed the first point of the fancurve to the same level as the second point (1400, 1400). This method is implemented now in function `wmi_read_fancurve_idx`. FanId and SensorId parameters doesnt make any difference. 

There are many Fan Curves defined and can be queried with the following powershell :
```powershell
$classInstance = Get-CimInstance -Namespace root/wmi -ClassName LENOVO_FAN_TABLE_DATA
$classInstance
```

```text
Active                      : True
CurrentFanMaxSpeed          : 3900
CurrentFanMinSpeed          : 0
DesignMaxFanSpeedNumber     : 10
EndOnlyUpwardAdjustNumber   : 10
Fan_Id                      : 1
FanSpeedStep                : 100
FanTable_Data               : {1400, 1700, 1900, 2200, 2500, 2700, 2900, 3100, 3500, 3900}
FanTable_Len                : 10
InstanceName                : ACPI\PNP0C14\GMZN_0
[...]

PR #571, opened 2026-09-14
feat(kernel): dynamic per-fan level RPM ladders from LENOVO_FAN_TABLE_DATA; drop hardcoded LEVEL_FAN_TABLES

st0nie:
Implements the dynamic alternative suggested by @qquique in the #567 review and drafted in #509, replacing the hardcoded per-model `LEVEL_FAN_TABLES` that #567/#568 added to `legion.py` with data read live from the firmware's `LENOVO_FAN_TABLE_DATA` WMI block.

## Kernel (`model_kwcn`, `model_q7cn`, new `has_fantable_data` flag)

- Query the data block (GUID `87FB2A6D-D802-48E7-9208-4576C5F5C8D8`, the thing DSDTs expose as `GZFD.WQA3`/`WQA7`/...) by row index until it runs out of entries.
- Row layout per #509's `struct WMIFanTableDefaultData` (validated on LZCN) and the fields Lenovo Legion Toolkit reads via WQL; variable-size arrays are validated against the returned buffer length, and rows that do not parse into a strictly ascending ladder are rejected.
- Keep the rows for the fan/sensor pairs LLT documents — (fan 1, 0x04) and (fan 2, 0x05) — prefer the current power mode's row (standard modes carry their own id; Legion Zone numbers custom as 0x100 and extreme as 0xE1; first row per fan is the fallback), and cache the ladders in `legion_private`, refreshed whenever `current_powermode` changes.
- Expose them as `fan1_level_rpm_table` / `fan2_level_rpm_table` on the platform device, e.g. `1700 1900 2100 2300 2500 2900 3400 3700 4400 5400`. Absent/unparseable block → attribute returns -ENODATA.

## Python

- `LEVEL_FAN_TABLES` and the BIOS-prefix lookup are gone. `FanCurveIO` reads the ladders from the platform device sysfs, so the values always match the current power mode and a new level-index model needs no userspace change at all.
- Degradation: older module / model without the block → previous `fan1_max`-based linear conversion. `LEVEL_POINT_MIN` stays as the fixed cross-firmware policy matching LLT's `GetMinimumFanTableAsync`.
- New `tests/test_fan_levels.py` (18 tests): sysfs reader, RPM↔level conversion (nearest, ties to lower, point-minimum clamps), FanCurveIO wiring.

## Verification

- `tests/test_kernel_checkpath.sh`: 0 errors / 0 warnings
- `tests/test_python.sh\': pylint 10.00/10; `black --check` clean; `tests/test_python_cli.sh` passes; new unit tests pass.
- **Needs hardware validation** (I have no Legion WMI3-fancurve machine): on a KWCN/Q7CN unit, `cat /sys/bus/platform/drivers/legion/legion/fan1_level_rpm_table` should print the ladder from the PR descriptions (KWCN fan 1: 1700..5400) and `dmesg` should show `fan table data: fan 1 has 10 levels, 1700..5400 RPM`. The layout decode is validated against #509's LZCN row and LLT's field list; a parse failure logs once and simply leaves the attributes at -ENODATA, so the risk of landing this before hardware confirmation is a no-op fallback, not a wrong table.

Refs: #509 (draft), #567 (review), #570 (context).


qquique:
- `FANTABLE_MAX_LEVELS 16` should be 10 as default

- if you use this [function](https://github.com/johnfanv2/LenovoLegionLinux/pull/347/changes/ef251a34be9624bdfee92facf51b9d9db2488670#diff-44eead4b16ced2831f1b7f5730357166c5160d107adeff9cee6b23626c6ddab2R1655). you dont need this : 

```c
#define FANTABLE_ROW_HEADER_SIZE (2 + 2 + 4)
#define FANTABLE_ROW_MIDDLE_SIZE (4 + 4)
#define FANTABLE_ROW_TAIL_SIZE (1 + 1 + 2 + 1 + 1 + 2 + 2 + 2 + 2 + 2)
```
> maybe thats why the AI used 16 for levels, it was trying to reach the positions after the headers of the acpi_object. 

- `.has_fantable_data` is the same concept as `.has_fancurve_defaults`, not the implementation, just means that if there is default data of the fans in the firmware, just one flag could be used.

- `FANTABLE_MAX_ROWS 32`, the table should be dynamic, the minimun is `powermodes (5) * fans (cpu,gpu,ic = 3) = 15`, q8cn has more than 50 I think it depends of status of the laptop if is in (AC/DC,Hybrid Mode,etc) , use instead `wmi_instance_count(GUID)` to obtain the number of rows.

- as mentioned in #485 all powerlimits of "other method features" should use the Capability Data GUID's, and there was a PR that added something missing the Discrete Data GUID's, no need for access method ACCESS_METHOD_WMI3_CLAMPED, the model that used it only needed to use its `.has_pl_coupling` flag for two features. And the gpu_oc (for the clamped code path) I still think is incorrect it should be using gpu_ppab.

- no need for `.fan_max_rpm = 5400,` it should be from the fan table field and replace the global MAX_RPM and the sysfs fan_maxspeed.

- `priv->current_powermode = powermode;` this only works if manually we modify the sysfs, but not when Fn+q or platform profile is modified, needs more places to be updated with getsmartfan method.

- `.fanfullspeed_requires_custom_powermode = true,` not needed, also the field `capability` in the capabilities data tables for that feature if in WMI3 access mode is set to allow (0x7, 0x407) or not, so the function `wmi_write_fanfullspeed_other` also needs to use the `common_store`.

st0nie:
@qquique thanks for the detailed review — all six points are implemented in the follow-up PR #573:

1. `FANTABLE_MAX_LEVELS` is 10 now, and rows are parsed with the `wmi_exec_query_ints()` pattern from #347 (exact buffer-length check) — the manual header/middle/tail size math is gone.
2. Row count comes from `wmi_instance_count()` instead of a fixed 32-row scan (50+ rows on models like Q8CN).
3. `has_fantable_data` is dropped; the ladders ride on `has_fancurve_defaults` (same concept, as you said), additionally gated on WMI3 fancurve access and the WMI block being present. kwcn/q7cn flip it on.
4. `ACCESS_METHOD_WMI3_CLAMPED` is removed: clamping is folded into `wmi_common_method_other_store()` (which also resolves its capability-data TODO); kwcn/secn/q7cn use plain WMI3, PL1/PL2 coupling stays behind `has_pl_coupling`, and the four stores that only had a CLAMPED branch (cpu_temperature_limit, cpu_l1_tau, gpu_power_target_offset, gpu_temperature_limit) gained the WMI3 branch so nothing regresses.
5. `gpu_oc`'s duplicate Other-Method mapping is dropped — 0x0201 belongs to `gpu_ppab_powerlimit`; where `LENOVO_GPU_METHOD` is absent, `gpu_oc` stays hidden via `legion_attribute_uses_gpu_wmi()`, so models with a real GPU method (lpcn/lzcn/...) keep working.
6. `fan_max_rpm` hardcoding removed from kwcn/q7cn; `fan1_max`/`fan2_max` now report the row's `current_fan_max_speed` per fan.

Local build warning-free, checkpatch 0/0. CI is running on #573.

PR #573, opened 2026-09-14
fix(kernel): address #571 review - exact-size fan table parsing, dynamic rows, one flag, WMI3 clamp fold, table-derived fan max

st0nie:
Follow-up to #571 implementing @qquique's review:

1. **`FANTABLE_MAX_LEVELS` = 10** — firmware tables always have 10 levels (LLT rejects other lengths).
2. **Exact-size row parsing** — new `wmi_exec_query_ints()` (the #347 pattern: `acpi_process_buffer_to_ints` validates the buffer length against the struct), so the manual `FANTABLE_ROW_HEADER/MIDDLE/TAIL_SIZE` arithmetic and memcpy bounds are gone; a row whose buffer isn't exactly the 10-level struct is rejected up front.
3. **One flag** — `has_fantable_data` removed; the ladders ride on `has_fancurve_defaults` (same concept: the firmware has default fan data), additionally gated on WMI3 fancurve access + the WMI block being present. `model_kwcn`/`model_q7cn` flip `has_fancurve_defaults` on: the defaults store writes a 1..10 table, which is exactly the level table these models take. `fancurve_defaults_powermode` becomes visible on them (harmless on q7cn: Fan_Set_Table ignores the mode byte and the level payload is valid).
4. **Dynamic row count** — `wmi_instance_count(GUID)` replaces the fixed 32-row scan (Q8CN has 50+ rows per AC/DC/hybrid state).
5. **`ACCESS_METHOD_WMI3_CLAMPED` removed** — `wmi_common_method_other_store()` now always clamps writes via the capability data (this resolves its own long-standing TODO); no capdata row → passthrough, so behavior on non-capdata models is unchanged. kwcn/secn/q7cn use plain WMI3; PL1/PL2 coupling stays driven by `has_pl_coupling`; the four stores that only had a CLAMPED branch (`cpu_temperature_limit`, `cpu_l1_tau`, `gpu_power_target_offset`, `gpu_temperature_limit`) gained the WMI3 branch so nothing regresses on the former CLAMPED models. `gpu_oc`'s duplicate Other-Method mapping is dropped: 0x0201 is `gpu_ppab_powerlimit`, and `gpu_oc` stays hidden where `LENOVO_GPU_METHOD` is absent.
6. **Fan max from the table** — `fan1_max`/`fan2_max` report the row's `current_fan_max_speed` (per-fan now) instead of hardcoded `fan_max_rpm` on kwcn/q7cn; `MAX_RPM` remains the fallback for models without the block.

Verified: local build (6.6 headers) warning-free, checkpatch 0/0. Not hardware-tested — same fallback-safe profile as #571 (no table → attrs return -ENODATA / config fallback).


qquique:
- 	in function `fan_max_show`, the condition should be only if it .`has_fancurve_defaults` = true

```c
if (priv->conf->has_fancurve_defaults &&
	    priv->conf->access_method_fancurve == ACCESS_METHOD_WMI3) {
```

because if the lapptop model accepts read/write for the fancurve using other access method like EC/other, if the machine has fantable defaults means that the data in the firmware can be obtained via wmi. 


changing that, on my LZCN :

```bash
$ cat fan1_level_rpm_table
1400 1700 1900 2200 2500 2700 2900 3100 3500 3900
$ cat fan2_level_rpm_table
1400 1700 1900 2200 2500 2700 2900 3100 3500 3900
$ cat fan1_max
3900
$ cat fan2_max
3900

$ sensors
legion_hwmon-isa-0000
Adapter: ISA adapter
Fan 1:           1400 RPM  (max = 3900 RPM)
Fan 2:           1400 RPM  (max = 3900 RPM)
CPU Temperature:  +41.0°C
GPU Temperature:  +35.0°C
IC Temperature:    +0.0°C

```

and the wmi custom read function stills shows the index, so the python client can use fan?_level_rpm_table 

```
u(speed_of_unit)|speed1[u]|speed2[u]|speed1[pwm]|speed2[pwm]|acceleration|deceleration|cpu_min_temp|cpu_max_temp|gpu_min_temp|gpu_max_temp|ic_min_temp|ic_max_temp
1	 1	 1	 2	 2	 0	 0	 0	 0	 0	 0	 0	 0
1	 2	 2	 5	 5	 0	 0	 0	 0	 0	 0	 0	 0
1	 3	 3	 7	 7	 0	 0	 0	 0	 0	 0	 0	 0
1	 4	 4	 10	 10	 0	 0	 0	 0	 0	 0	 0	 0
1	 5	 5	 12	 12	 0	 0	 0	 0	 0	 0	 0	 0
1	 6	 6	 15	 15	 0	 0	 0	 0	 0	 0	 0	 0
1	 7	 7	 17	 17	 0	 0	 0	 0	 0	 0	 0	 0
1	 8	 8	 20	 20	 0	 0	 0	 0	 0	 0	 0	 0
1	 9	 9	 22	 22	 0	 0	 0	 0	 0	 0	 0	 0
1	 10	 10	 25	 25	 0	 0	 0	 0	 0	 0	 0	 0
```

and instead of showing 1,2,3,4,5,6,7,8,9,10 , can mask them with the values from fan?_level_rpm_table, and  the user could change the order and then submitting that order with wmi custom write function. 

to take in consideration : 
- the values of the fantable data sensor_temp[FAN_TABLE_MAX_LEVELS] can be (depends on the firmware) max temps for cpu, gpu, ic , so at some point would be good to have also them included in the fan_level_rpm_table or other sysfs table, to show them on the client too.

edit: 
still needs to refresh the fan table when changing powermode with fn + q / platform profile
--- I found why 
it is refreshing (only with sysfs powermode), but somehow the code looks like is from other project 
this function `fantable_row_to_ladder` is checking if the first level rpm is 0 , and then reduces the levels of the fancurve, and skips the first one,  looks like is more like a validation to save the curve instead of reading the curve, had to change the function to  : 

```c
static bool fantable_row_to_ladder(const struct wmi_fantable_row *row,
				   struct fantable_ladder *ladder)
{
	bool has_off_slot = false; // changed
	u8 count = row->fan_table_len;
	u8 i;

	if (count < 1 || count > FANTABLE_MAX_LEVELS)
		return false;

	for (i = 0; i < count; i++) {
		u16 rpm = has_off_slot ? row->fan_speed[i + 1] :
					 row->fan_speed[i];

	// 	if (rpm < 1)
	// 		return false;
		if (i > 0 && rpm <= ladder->rpms[i - 1])
			return false;
		ladder->rpms[i] = rpm;
	}

	ladder->level_count = count;
	ladder->max_rpm = row->current_fan_max_speed;
	if (!ladder->max_rpm)
		ladder->max_rpm = ladder->rpms[count - 1];
	return true;
}

```
now I can see my fancurves correctly : 
```c
$ cat powermode
2
$ cat fan1_level_rpm_table
0 1400 1700 1900 2200 2500 2900 3100 3500 3900
$ cat powermode
3
$ cat fan1_level_rpm_table
1400 1700 1900 2200 2500 2700 2900 3100 3500 3900
```

- needs to update `priv->current_powermode` in functions `write_powermode` and `legion_wmi_notify`

- and last one for today , same as first pointl: 

```
	if ((attr == &dev_attr_fan1_level_rpm_table.attr ||
	     attr == &dev_attr_fan2_level_rpm_table.attr) &&
 !(priv->conf->has_fancurve_defaults && //
// -----> 	      priv->conf->access_method_fancurve == ACCESS_METHOD_WMI3 &&
	      wmi_has_guid(WMI_GUID_LENOVO_FANTABLE_DATA)))
		return 0;

```

PR #577, opened 2026-09-14
fix(kernel): refresh the cached power mode before selecting per-mode firmware data

Hishammm0:
Model: **Lenovo Legion Pro 5 16IRX8 (82WK), BIOS KWCN54WW**, CachyOS 7.2.4-3-cachyos. The fix is not
model-specific.

Follow-up to #567, addressing the Copilot review comment on the README coexistence paragraph ("`current_powermode`
is initialized at probe and only updated in this driver's `powermode_store`; `ACCESS_METHOD_WMI3_CLAMPED` uses that
cached value, so the claimed own power-limit path can clamp writes against the previous mode after the mainline
driver changes profiles"). That comment was suppressed at review time but the observation is correct on `main`.

`priv->current_powermode` is set at probe and afterwards only by `powermode_store()`. The mode also changes without
it, in the default setup as well as the coexistence one:

- `legion_platform_profile_set()` writes the mode directly, so a profile change through
  `/sys/firmware/acpi/platform_profile` or power-profiles-daemon bypasses the cache with the default
  `enable_platformprofile=1`;
- with `enable_platformprofile=0` the mainline `lenovo-wmi-gamezone` driver owns the platform profile (the setup the
  README now recommends);
- Fn+Q goes through `lenovo-wmi-events` when both are loaded.

Two things key on the cached value:

- `clamped_value()` selects the `LENOVO_CAPABILITY_DATA_01` row by mode byte, so a power-limit write after a
  profile change is clamped against the previous mode's range. On the 16IRX8 every mode publishes the same ranges,
  which is why it was not visible in testing; the row lookup exists precisely for firmware where they differ.
- `fantable_ensure()` keeps the cached ladder as long as `fantable_powermode` matches the cache, and
  `fantable_refresh()` picks rows by it, so per-mode ladders would never be re-read after an external mode
  change.

The change adds `sync_powermode_locked()`, a single `GetSmartFanMode` read that updates the cache and keeps the old
value on error, and calls it where the cache is consumed: in `clamped_value()` under the fan-curve mutex (none of
its three callers hold it at that point), in `fantable_ensure()` (all callers already hold the mutex), and in
`powermode_show()`, which as a side effect now reports the cached value instead of an uninitialised `int` when the
read fails.

Cost: one `GetSmartFanMode` read per power-limit write, per level-table read, per `fan1_max`/`fan2_max` read (on the
`has_fancurve_defaults` + WMI3 models, where `fan_max_show()` goes through `fantable_ensure()`) and per `powermode`
read. It is the same class of WMI call as a `fan1_input` read, so a 1 Hz `sensors` poll roughly doubles the
fan-channel WMI traffic. If that matters, the sync can be memoised with a short jiffies TTL inside
`sync_powermode_locked()`; I kept it unconditional so the clamp is never a second behind a profile change.

## Verification

Built with plain `make` on CachyOS 7.2.4-3-cachyos (the clang auto-detection from #566), `tests/test_kernel_checkpath.sh`
0 errors / 0 warnings, clang-format clean on the touched lines.

On the machine, with the mainline drivers loaded and `legion_laptop enable_platformprofile=0`:

```text
$ echo custom | sudo tee /sys/class/platform-profile/platform-profile-0/profile   # mainline switches the mode
$ cat /sys/bus/platform/drivers/legion/legion/powermode                            # 255, as before (live read)
$ cat /sys/bus/platform/drivers/legion/legion/fan1_level_rpm_table                 # re-selected for mode 255
```

Behaviour is unchanged when `legion_laptop` owns the platform profile, since `powermode_store()` still updates
the cache directly.


Hishammm0:
Both Copilot points were right; pushed as an amended single commit (the change is part of the same fix):

- **`clamped_value()` TOCTOU:** the capability-row scan now compares against a snapshot of the mode taken under
  `fancurve_mutex` right after the sync, instead of re-reading `priv->current_powermode` after unlocking.
- **Unlocked cache write:** `powermode_store()` assigned `priv->current_powermode` after releasing the mutex; the
  assignment now happens under it, on success only, so every writer of the cache holds `fancurve_mutex`.

On the settling-window part of the suppressed comment: a sync that runs during the 500 ms after a mode write can
indeed read the previous mode, but it can only affect that one call. Every consumer syncs immediately before it
uses the value, so the cache is never trusted across calls; the next power-limit write, ladder read or `powermode`
read syncs again and sees the settled mode. Suppressing syncs while a write settles would need a flag plus a timer
for a window the firmware itself reports as the old mode, which did not seem worth it. Happy to add it if you
prefer.

Rebuilt on CachyOS 7.2.4-3-cachyos: plain `make` clean, `W=1` no new warnings, `tests/test_kernel_checkpath.sh`
0/0, clang-format clean on the touched lines.

armoox:
Hey, took a look at this — it's good stuff. The stale cache bug is real, I checked the code on main and `legion_platform_profile_set()` definitely writes the mode without touching `current_powermode`, so `fantable_ensure` and `clamped_value` can end up using old data after an external profile change. The sync + snapshot approach makes sense, and moving the cache write in `powermode_store` under the lock fixes that race too. The `powermode_show` fix returning the cached value instead of garbage on read failure is a nice bonus.

Looks good to merge.

One thing — if you haven't already, check the `AGENTS.md` in the repo root. There's some guidance on AI tool usage for this project. The TL;DR is don't lean on AI models too much for this kind of work — kernel driver code that touches real EC hardware is exactly the stuff where AI can confidently get things wrong. Totally fine to use it, just make sure you're double-checking everything yourself and not treating its output as gospel. With hardware code especially, you really want to be the one who understands what the patch does, not just trusting that it's right.

Hishammm0:
Thanks for the review, and for checking the cache path on main yourself.

Point taken on AGENTS.md. I did use Claude Code for this series, and I want to be straight about how: it drafted the
patches, but nothing went to the branch before I ran it on the 16IRX8 with the mainline drivers loaded (the custom
switch → `powermode` 255 → level 9 → 4400/4600 RPM run in the PR body), and every revision was pushed by me after
reading the diff. What the patch does, in my words: the driver cached the power mode and three paths could change
the mode without updating the cache, so the capability-row clamp and the ladder cache could key on old data; the
fix re-reads the mode from firmware wherever the cache is consumed, under the existing mutex, and snapshots it so a
concurrent store cannot race the row scan. I'll keep the hardware-touching parts on the short leash you describe.

armoox:
Thanks for the reply, and really — no offense intended at all, I hope it didn't come across that way. I completely understand using Claude Code or any AI tool to speed things up, that's totally fair game.

What I was getting at is: use AI to help you *understand* the project — its architecture, the WMI/EC flows, how the drivers fit together. That's genuinely useful. Just be careful about blindly accepting code it generates, especially with hardware-touching stuff like this. You've already shown the right approach though — you ran it on your actual machine, read the diffs before pushing, and you clearly understand what the patch does. That's exactly the balance I was hoping for.

I wasn't criticizing how you worked, just wanted to make sure it stays that way — the EC is not very forgiving if something gets merged that nobody fully understood. Keep doing what you're doing, the PRs are good.

Hishammm0:
No offense taken at all, promise :) Your point is completely valid, and honestly it's the kind of reminder I'd
rather get once too often than once too few. An EC that holds a grudge is a good reason to read every diff twice,
and I do check the code before it goes anywhere, to the best of my ability; the AI drafts, I read, I run it on the
actual laptop, and only then it gets pushed.

I'm just happy to be working on this in my free time while I migrate to CachyOS for good. Having my fans behave
the way they did on Windows, and then a bit better, has been a very satisfying way to learn how the WMI/EC side of
this laptop actually works. Thanks for the review and the kind words!

PR #578, opened 2026-09-14
fix(kernel): hide fan_maxspeed where the firmware lacks the max-speed methods (KWCN, Q7CN, RLCN)

Hishammm0:
Model: **Lenovo Legion Pro 5 16IRX8 (82WK), BIOS KWCN54WW**, CachyOS 7.2.4-3-cachyos.

Follow-up to #567, addressing the suppressed Copilot comment on the `model_kwcn` configuration: `fan_maxspeed` calls
`Fan_Get_MaxSpeed` / `Fan_Set_MaxSpeed` (LENOVO_FAN_METHOD ids 3/4), but KWCN54WW implements only `Fan_Get_Table`
and `Fan_Set_Table` on that GUID (the Windows MOF lists only those two). The DSDT's `WMAB` handler has no branch for
ids 3/4 and falls through, so on this machine the attribute reads an implicit `0` and a write touches nothing:

```text
$ cat /sys/bus/platform/drivers/legion/legion/fan_maxspeed
0
$ journalctl -k | grep simple_wmi
legion_laptop: get_simple_wmi_attributewith raw value: 0
```

The Windows dump of the same firmware points the same way for the related interfaces: the GameZone `GetFanMaxSpeed`
getter (a different GUID) fails with `Invalid object` (HRESULT 0x8004100f) and `LENOVO_FAN_MAX_SPEED_DATA`
enumerates as `Not supported`.

The change adds `skip_fan_maxspeed` to `struct model_config`, following `skip_lockfancontroller`, sets it for
`model_kwcn` and hides the attribute in `legion_sysfs_is_visible()`. The per-fan maximum for these models is already
reported by hwmon `fan1_max` from `LENOVO_FAN_TABLE_DATA` (#571). The Python tools and `legiond` do not use
`fan_maxspeed`; `fan_control.sh` did write it unconditionally after `fan_fullspeed` and would have aborted with a
misleading error on these models, so it now skips the attribute when the driver does not expose it. The Q7CN and
RLCN header comments in the file record the same Fan Method surface (`WMAB` implements only ids 5/6), so
`model_q7cn` and `model_rlcn` get the flag too; their owners can confirm on hardware. Both READMEs and
`doc/FEATURES_AND_TESTING.md` now say which models hide the attribute and why.

## Verification

Built with plain `make` on CachyOS 7.2.4-3-cachyos, `tests/test_kernel_checkpath.sh` 0 errors / 0 warnings,
clang-format clean on the touched lines, `bash -n` on the script. On the machine, with the patch loaded,
`fan_maxspeed` is no longer listed under the platform device; every other attribute, the fan ladders and `fan1_max`
are unchanged, and `./fan_control.sh status` / `disable` complete normally.


Hishammm0:
Both points taken, pushed as an amended commit:

- **Q7CN / RLCN:** their header comments in `legion-laptop.c` record the same Fan Method surface (`WMAB` implements
  only `Fan_Get_Table(5)` / `Fan_Set_Table(6)`), so `model_q7cn` and `model_rlcn` now set `skip_fan_maxspeed` as
  well. I could only verify KWCN on hardware; on those two the firmware does not implement ids 3/4 either, so the
  attribute cannot work (it would read an implicit value or fail) and hiding it is the safe direction. @st0nie /
  the RLCN owner can confirm.
- **Docs:** the Legion Zone v3 paragraph in `README.md` and `README_zh-hans.md`, and the level-index section of
  `doc/FEATURES_AND_TESTING.md`, now state that `fan_maxspeed` is not exposed on KWCN / Q7CN / RLCN and that
  `fan_control.sh` skips it.

Rebuilt on CachyOS 7.2.4-3-cachyos: `make` clean, `tests/test_kernel_checkpath.sh` 0/0, clang-format clean on the
touched lines; the amended module is what this 16IRX8 runs via DKMS.

armoox:
Hey, looked this over — looks solid. The KWCN/Q7CN/RLCN firmware limitation is already documented in the model comments (they all say WMAB only does ids 5/6), so hiding `fan_maxspeed` makes total sense. The flag follows the same pattern as `skip_lockfancontroller`, the sysfs visibility gate is clean, and the `fan_control.sh` fix is a nice catch — that script was definitely going to blow up on these models. Docs look good across both READMEs and FEATURES_AND_TESTING.

Looks good to merge.

One thing — if you haven't already, check the `AGENTS.md` in the repo root. There's some guidance on AI tool usage for this project. The TL;DR is don't lean on AI models too much for this kind of work — kernel driver code that touches real EC hardware is exactly the stuff where AI can confidently get things wrong. Totally fine to use it, just make sure you're double-checking everything yourself and not treating its output as gospel. With hardware code especially, you really want to be the one who understands what the patch does, not just trusting that it's right.

Hishammm0:
Thanks. Same note as on #577 about AGENTS.md: understood, and the hardware side of this one is verified on my
16IRX8 (attribute gone after the DKMS reinstall, everything else unchanged, `fan_control.sh status` completes).
For Q7CN/RLCN I only have the DSDT analyses recorded in the model comments, so the flag there is inferred from
the same `WMAB` shape; if either owner sees `fan_maxspeed` doing anything real on their unit, the flag should come
off for that model and I'll follow up.

armoox:
Thanks for the reply, and really — no offense intended at all, I hope it didn't come across that way. I completely understand using Claude Code or any AI tool to speed things up, that's totally fair game.

What I was getting at is: use AI to help you *understand* the project — its architecture, the WMI/EC flows, how the drivers fit together. That's genuinely useful. Just be careful about blindly accepting code it generates, especially with hardware-touching stuff like this. You've already shown the right approach though — you verified the attribute behavior on your actual machine and tied the Q7CN/RLCN reasoning to the DSDT comments, giving owners a clear path to correct you if the inference is wrong. That's exactly the balance I was hoping for.

I wasn't criticizing how you worked, just wanted to make sure it stays that way — the EC is not very forgiving if something gets merged that nobody fully understood. Keep doing what you're doing, the PRs are good.

Hishammm0:
All good here, no offense taken, and the reminder is welcome :) Same as I said over on #577: I read the diffs and
run them on the real machine before pushing, to the best of my ability, and for the models I can't test I'd rather
leave the owners an easy way to prove me wrong than pretend certainty.

This has been a fun free-time project while I move to CachyOS full time, so thank you for the thorough looks and
the patience with the back and forth. On to the next one!

Issue #529, opened 2026-09-03
[BUG] Crash on attempt to change acc./dec. time to values not from 2 to 5


HammerPoppy:
**Problem Description**
Both legion_gui and legion_cli crashes when trying to write to HW Fan curve with acceleration or deceleration time values that is not from range 2-5 (2 or 3 or 4 or 5). Last couple of output strings:
```bash
  File "/usr/lib/python3.14/site-packages/legion_linux/legion.py", line 827, in _write_file
    with open(file_path, "w", encoding=DEFAULT_ENCODING) as filepointer:
         ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 95] Operation not supported
```
  
Has the bug occurred following steps described in the README?
Yes and No. Steps were done, apps work great at basic level, except for when trying to change acc./dec. time to "wrong" values.
  
What steps did you take?
Tried installing app from compiling git repo, then installed it from cachyos/lenovolegionlinux-dkms to install legiond automatically. A lot of configuring and changing configs that I can't recall now. Then I accidentally figured that apps crash when trying to write to HW acc./dec. time values that is not from 2-5.
  
What commands did you enter?
```bash
sudo legion_gui
```
then "Read from HW", then change any value in Acceleration Time [s] or Deceleration Time [s] to for example 1 and "Apply to HW"
  
also
  
```bash
sudo legion_cli fancurve-write-preset-to-hw balanced-ac
```
when balanced-ac profile has acc./dec. time value/s for example 1

What was the (full) error you received and what command or action did lead to it?
legion_gui:
```bash
OSError: [Errno 95] Operation not supported

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/legion_linux/legion_gui.py", line 871, in on_write_fan_curve_to_hw
    self.model.write_fancurve_to_hw()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/legion_linux/legion.py", line 1666, in write_fancurve_to_hw
    self.fancurve_io.write_fan_curve(self.fan_curve, write_minifancurve)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #580, opened 2026-09-15
fix(gui): allow loading presets without prior view validation and repair profile accel/decel

### LINKED ISSUE #579: [BUG] legion_gui: 'Load from Preset' blocked by validation error on uninitialized view (Regression from 112dfef)
**Problem Description**
After upgrading to the latest commit (`112dfef` / PR #529), opening `legion_gui` and clicking "Load from Preset" fails with the warning popup:
> *"Invalid value in fan curve point 1: acceleration and deceleration time must be between 2 and 5 (the fan controller rejects other values)."*

This prevents loading any preset (including valid presets like `balanced-performance-ac` and `performance-ac`).

**What steps did you take?**
1. Boot system (the hardware fan curve initial acceleration/deceleration values returned from EC default to 0).
2. Launch `legion_gui`.
3. Select a preset in the "Fancurve Preset" dropdown (e.g., `balanced-performance-ac`).
4. Click "Load from Preset".

**What was the (full) error you received?**
`Invalid value in fan curve point 1: acceleration and deceleration time must be between 2 and 5 (the fan controller rejects other values).`

**Root Cause:**
1. In `python/legion_linux/legion_linux/legion_gui.py`, `on_load_from_preset()` unnecessarily validates the *currently displayed view* before loading the preset:
```python
    def on_load_from_preset(self):
        name = self.view_fancurve.preset_combobox.currentText()
        if not self._read_fancurve_from_view():  # <-- Redundant check
            return
        try:
            self.model.load_fancurve_from_preset(name)
```
When loading from a preset, the intent is to overwrite the GUI with values loaded from disk. Validating the current view (which holds hardware defaults where `accel == 0`) blocks the load operation completely.

2. Additionally, several shipped profiles in `extra/service/profiles/` (`quiet-battery.yaml`, `quiet-ac.yaml`, `balanced-ac.yaml`, `balanced-battery.yaml`, `performance-battery.yaml`) contain `acceleration: 0` or `1` in points 9 and 10, which will also trigger the new `2 <= accel <= 5` validation when written to hardware.

**Model and Debug Info**
* **Installation type:** AUR (`lenovolegionlinux-git` & `lenovolegionlinux-dkms-git` @ 2f26060)
* **Distribution:** CachyOS Linux (Kernel: 7.2.5-1-cachyos)
* **Product Name:** 82Y9
* **Product Version:** Legion R7000P APH8 (Lenovo Legion Slim 5 16APH8)
* **BIOS version:** M3CN43WW
* **CPU model:** AMD Ryzen 7 7840H w/ Radeon 780M Graphics
* **GPU model:** NVIDIA GeForce RTX 4060 Mobile / Max-Q + AMD Radeon 780M
* **Keyboard backlight:** 4-Zone RGB

conqazht:
### Summary of Changes

1. **Remove redundant view validation in `on_load_from_preset()`**:
   - In `legion_gui.py`, `on_load_from_preset()` called `_read_fancurve_from_view()` before loading from disk.
   - When launching `legion_gui`, the hardware EC fan curve acceleration/deceleration values initially default to 0. Commit 112dfef (#529) added strict range checks (`2 <= accel <= 5`), causing `_read_fancurve_from_view()` to block users from loading any preset with a validation error popup.
   - Since loading a preset replaces the current view with data from disk, validating the current view before loading is unnecessary and prevents loading presets on uninitialized views.

2. **Repair points 9 & 10 in shipped profiles**:
   - `balanced-ac.yaml`, `balanced-battery.yaml`, `performance-ac.yaml`, `performance-battery.yaml`, `quiet-ac.yaml`, and `quiet-battery.yaml` had acceleration and/or deceleration set to 0 or 1 in points 9 and 10.
   - These are updated to `2` to comply with the controller's range constraint.

3. **Add unit test**:
   - Added `test_load_from_preset_with_uninitialized_view` to `tests/test_gui_startup.py` to ensure loading presets works as expected when the view contains uninitialized entries.

Fixes #579


armoox:
Reviewing this from the side of the fan-curve trim logic this PR interacts with.

**The load-path change is correct.** Validating the currently displayed view before `on_load_from_preset()` overwrites it from disk is redundant, and it's exactly what blocks loading presets on hardware whose EC boots with accel/decel = 0 (per #579). No objection there.

**The preset changes are only partially correct.** Updating accel/decel to 2 on *real* curve points is fine and necessary — e.g. `performance-*` points 9–10, and `balanced-ac`/`balanced-battery` point 9 (fan 4000, real temps), which truly had `accel: 1` and need to satisfy the controller's 2–5 range.

The problem is setting accel/decel = 2 on the **all-zero trailing entries**:
- `balanced-ac` point 10, `balanced-battery` point 10, `quiet-ac` points 9+10, and `quiet-battery` points 9+10 are entirely zero (`fan1_speed: 0`, all temps `0`).
- `FanCurveEntry.is_empty()` (added in #493) exists specifically to recognize these as EC padding and ensure they are **never written to hardware**. #493 also trims them on read, and `write_fan_curve` trims to the live `auto_points_size` (#511).
- Once accel/decel become 2, `is_empty()` returns `False`, so these padding points are no longer classified as padding and are no longer trimmed. On any model/mode where the EC supports a full 10 points (e.g. performance/balanced AC), points 9–10 now get written as *real* entries (0 RPM, 0 °C, accel 2). This directly contradicts the convention the repo adopted in #493 ("trailing all-zero entries are never meaningful") and risks reintroducing the `EOPNOTSUPP` write failures that #511 fixed, since hardware may reject a zero-fan/zero-temp point written as a regular entry.

Suggested alternative that keeps both fixes consistent:
1. Keep the removal of `_read_fancurve_from_view()` from `on_load_from_preset()` — agreed.
2. Instead of rewriting padding to 2, trim first and validate after: drop trailing empty entries in `write_fancurve_to_hw()` (or in `load_fancurve_from_preset()`), then run the accel/decel range check on whatever remains. That way shipped-preset padding stays harmless and validation can never trip on entries that are never written.

st0nie:
Thanks for the careful review — addressed in the latest push (31e0ad5):

1. **Padding entries reverted to all-zero.** `balanced-*` point 10 and `quiet-*` points 9-10 are back to `acceleration/deceleration: 0`, so `FanCurveEntry.is_empty()` (#493) classifies them as padding again and `FanCurveIO.write_fan_curve` keeps trimming them (#511). The `1 -> 2` repairs on *real* points (`performance-*` 9-10, `balanced-*` 9) stay, since the controller's 2-5 range genuinely applies there.

2. **Trim first, validate after — at the validation site.** The accel/decel range check moved from `FanCurveEntryView.get()` to `FanCurveTab.get_fancurve()`: trailing empty entries are dropped for validation only, and the 2-5 check runs on the writable prefix. The returned `FanCurve` keeps all entries, so save-to-preset semantics are unchanged, and `write_fan_curve` still does the actual trim. A mid-curve all-zero entry is *not* trailing, so it still gets validated (and would still fail on accel=0) — same conservatism as before for anything that could actually be written.

3. **Tests added** for both directions: a curve with real points + trailing padding no longer blocks "Apply to HW", and a real point with accel=1 still does.

The load-path change (`on_load_from_preset` no longer validating the view) is unchanged, per your assessment.

st0nie:
Follow-up for the record: the lint failure this PR hit after the review-fix commit was caused by the fix commit accidentally carrying an 88-column reformat (a `black` run that missed `python/legion_linux/pyproject.toml`), which also displaced three `# pylint: disable=unused-argument` comments and re-triggered W0613. Cleaned up on main in 56bdc0a — the semantic content of the review fix (padding stays all-zero, trim-before-validate) is unchanged. Thanks again to @conqazht for the PR and to @armoox for the review that caught the padding issue.

PR #583, opened 2026-09-16
fix(gui): catch RuntimeError on fancurve write failure to prevent crash

conqazht:
## Description

When running `legion_gui` with `--use_legion_cli_to_write` (the default mode), failures during `write_fancurve_to_hw()` raise a `RuntimeError` via `write_file_with_legion_cli()` (`legion.py:169`):
```python
raise RuntimeError(f"legion_cli failed for {name} with code {returncode}: {err_str}")
```
Previously, `on_write_fan_curve_to_hw()` only caught `OSError`. Consequently, any failure from `legion_cli` (such as a hardware write rejection, ACPI failure, or cancelled pkexec prompt) caused an unhandled `RuntimeError` that reached the Qt event loop, resulting in a `SIGABRT` crash instead of a warning dialog.

This PR:
- Catches both `(OSError, RuntimeError)` in `on_write_fan_curve_to_hw()` so that a `QMessageBox.warning` is properly shown to the user.
- Adds a unit test `test_write_handles_runtime_error` in `test_gui_startup.py` to ensure `RuntimeError` is handled gracefully without crashing.


armoox (commented):
@conqazht The fix correctly addresses the crash — `write_file_with_legion_cli()` (legion.py:169) raises `RuntimeError`, which escaped the `OSError`-only catch, and the new test exercises that path (empty view → `get_fancurve()` returns a valid empty curve → the mock RuntimeError reaches the handler). Test suite passes 8/8 locally (I ran it on this branch).

One suggestion: `read_fancurve_from_hw()` and `update_fancurve_gui()` at legion_gui.py:918-919 run after the catch and are themselves unwrapped — a failing readback would crash the same way. Consider returning after the warning or wrapping the readback too. Otherwise LGTM.

PR #596, opened 2026-09-18
fix(kernel): skip unsupported lockfancontroller on model_m3cn_8227

conqazht:
## Description

On Lenovo Legion Slim 5 16APH8 / R7000P 2023 (`model_m3cn_8227`, EC `0x8227`), fan control is handled via WMI3 (`\_SB.GZFD.SFAN`).

The legacy EC byte at `0xC4AB` (`EXT_LOCKFANCONTROLLER`) defaults to `0x08` (`EC_LOCKFANCONTROLLER_ON`) on every boot. Because `model_m3cn_8227` was missing `.skip_lockfancontroller = true`:
1. `/sys/devices/platform/legion/lockfancontroller` reported `1` (true) on every boot/module load.
2. `legion_gui` showed the *"Lock fan controller…"* checkbox permanently checked on startup.
3. Users had to manually untick the checkbox every session, which was confusing since the lock has no functional effect.

### Evidence that 0xC4AB is not a functional fan lock on EC 0x8227

| Test | Result |
|------|--------|
| Byte value across quiet (1), balanced (2), performance (3) modes | Always `0x08` — stable, not mode-dependent |
| Write `0` then readback | Write accepted, readback returns `0` |
| Write `0` in custom mode (255) then readback | Write accepted, readback returns `0` |
| **Custom fan curve via WMI3 while byte = 0x08 ("locked")** | **Works normally** — fans accelerate/decelerate per the written curve |

The critical observation: **WMI3 fan speed control works regardless of the byte's value**. On older Legions (2020–2021) where this byte functions as a real lock, writing `EC_LOCKFANCONTROLLER_ON` freezes the fan controller and rejects speed changes. On EC 0x8227, the byte is writable but has no observable effect on fan behavior — it is vestigial.

Setting `.skip_lockfancontroller = true` hides this non-functional sysfs attribute, matching what `model_q7cn` and `model_rlcn` already do. The `legion_gui` checkbox is then properly disabled instead of displaying a misleading lock state.

## Verification
- Kernel module compiles cleanly (0 warnings, 0 errors).
- Tested on Lenovo Legion Slim 5 16APH8 (82Y9, BIOS M3CN43WW, EC 0x8227).
- Sysfs attribute `lockfancontroller` hidden after applying the change.
- `legion_gui` correctly disables the lock checkbox.


armoox:
Hi, not a maintainer here, just someone who noticed this and wanted to share some concerns before you/others decide on this PR — and I want to be upfront that I'm not 100% sure I'm right about any of it. Please take all of this as "worth double-checking" rather than a verdict.

**1. The "matches model_lpcn / model_q5cn" claim in the description doesn't match the current tree.**

As of upstream `main` (v0.0.30), only two configs set `skip_lockfancontroller`:

- `model_q7cn` (kernel_module/legion-laptop.c:1822)
- `model_rlcn` (kernel_module/legion-laptop.c:1911)

`model_lpcn` does not set it, and I couldn't find a `model_q5cn` at all (there's `model_q6cn`, `model_q8cn`, etc.). Maybe the description is based on a different fork/version, and then this concern doesn't apply — but against upstream main it reads as inaccurate, so it might be worth correcting so nobody greps it later as precedent.

**2. The evidence for "0xC4AB is not a lock register" might be circular.**

Your proof is that `0xC4AB` "naturally reads `0x08`" — but `0x08` is exactly the value this driver defines as `EC_LOCKFANCONTROLLER_ON` (legion-laptop.c:5347). So "it reads 0x08" explains the *symptom* (driver + GUI think the lock is on) but doesn't by itself prove the byte is unrelated to fan locking. I could easily be misreading the situation though — the value 8 being "ON" is just this driver's convention, so if you have other evidence the byte means something else on EC 0x8227, I'd genuinely like to see it.

Things that would strengthen the claim, if you have access to test:
- Dump `0xC4AB` while switching Fn+Q modes (quiet/balanced/performance/custom) — is it stable?
- Compare against a model where the lock is *documented* to work: does the same byte read 0 there when unlocked? (q7cn/rlcn aren't a good reference here — they skip the attribute for the same reason this PR does, so they'd plausibly read the same 8.)
- Rule out the alternative that the lock exists but needs a prerequisite before the write is accepted — e.g. `fan_fullspeed` on the same family is gated behind custom powermode (`fanfullspeed_requires_custom_powermode`). If that's the case here, "writes bounce back to 8" could mean a working feature is being hidden rather than an unsupported one. I have no idea whether that's true for your EC — just flagging it as a hypothesis.

**3. This byte is part of a shared register map.**

`model_m3cn_8227` uses `ec_register_offsets_v0`, the same map many models use (where `EXT_LOCKFANCONTROLLER = 0xC4AB`), so a misread probably wouldn't be unique to the 82Y9. Notably `model_lpcn` (EC 0x5507) is a near-twin of your config (same WMI3 speed-only fancurve, same no-extreme gating) and does *not* skip the attribute — if the misread is real, lpcn users might see the same falsely-ticked checkbox. I'm not saying the same fix should be applied there blindly — just that it's worth asking whether lpcn is affected too. I don't have one to test, so I can't verify.

**4. Documentation side-note.**

If something like this lands, the project convention is that user-visible behavior changes update `README.md`, `README_zh-hans.md` and, if a test procedure changes, `doc/FEATURES_AND_TESTING.md`. Hiding an attribute + the GUI behavior change would seem to qualify. If the maintainers don't require that for a change this small, fine — I just wanted to flag it.

Again — I'm genuinely not certain my reading is right; the two-line change itself is clearly in the safe direction (stop writing to a byte that arguably shouldn't be poked), so I'm not arguing against it. I just think the description overclaims and the core claim could use one more piece of evidence. Happy to be corrected.

conqazht:
Thanks for the thorough review. You raised valid points and I've updated the PR description accordingly.

**1. Incorrect model references**: Fixed. The description now correctly references `model_q7cn` and `model_rlcn` as the two models that set `skip_lockfancontroller` on upstream main. `model_lpcn` does not set it and `model_q5cn` does not exist, my mistake.

**2. Evidence for the byte being non-functional**: You were right that "reads 0x08" alone is circular. I ran additional tests on real hardware (82Y9, BIOS M3CN43WW, EC 0x8227):

| Test | Result |
|------|--------|
| Byte value across quiet/balanced/performance modes | Always `0x08`, stable |
| Write `0` then readback | Write accepted, reads `0` |
| Write `0` in custom mode (255) | Write accepted, reads `0` |
| **WMI3 fan curve write while byte = 0x08 ("locked")** | **Works normally**, fans follow the written curve |

The byte is writable (correcting my earlier claim), but it has **no observable effect on fan behavior**. WMI3 fan control (`\_SB.GZFD.SFAN`) bypasses this legacy EC byte entirely. On older 2020-2021 Legions where the lock is functional, `0x08` freezes the fan controller and rejects speed changes. That does not happen here.

**3. model_lpcn**: Noted. I don't have that hardware so I can't confirm whether it's affected. Leaving it out of scope for this PR.

**4. Documentation**: Will defer to maintainers on whether a two-line attribute-hide warrants a docs update.

